PR #1, opened 2021-11-05
Add msw

evulins:
Mock Service Worker server has been added for the development environment. Setup was made based on MSW documentation -> https://mswjs.io/


src/mocks/browser.tsx
```diff
@@ -0,0 +1,6 @@
+// src/mocks/browser.js
```
franiakat: In plain Typescript files without any JSX elements we don't need to support it, so the file extension should be `browser.tsx`. ;)

src/mocks/handlers.tsx
```diff
@@ -0,0 +1,37 @@
+// src/mocks/handlers.js
```
franiakat: In plain Typescript files without any JSX elements we don't need to support it, so the file extension should be `browser.tsx`.

src/mocks/handlers.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+    // If authenticated, return a mocked user details
+    return res(
+      ctx.status(200),
+      ctx.json({
+        username: 'admin',
+      }),
+    )
+  }),
+]
```
franiakat: Empty line at the end is missing - in the future prettier will do this for you. 🙃

franiakat (changes requested):
Congrats on your first PR 🍾 @evulins & @bazelajoanna 🎉 
Please see the comments.

franiakat:
One additional thing - it is a good practice to add some description in the PR, so it can be used as a documentation in the future.
It would be helpful to say e.g. why given PR was created, or based on which sources (some links to documentation, designs - depending on the PR matter). Please, add one. :)

src/mocks/browser.js
```diff
@@ -0,0 +1,6 @@
+// src/mocks/browser.js
+import { setupWorker } from 'msw'
+import { handlers } from './handlers'
+
+// This configures a Service Worker with the given request handlers.
+export const worker = setupWorker(...handlers)
```
franiakat: Please, add empty line at the end - didn't spot it before.

package.json
```diff
@@ -40,7 +40,11 @@
...
   "devDependencies": {
+    "msw": "^0.35.0",
     "ts-loader": "^9.2.6",
     "typescript": "^4.4.4"
+  },
+  "msw": {
+    "workerDirectory": "public"
   }
-}
+}
```
franiakat, line 50: Please, add empty line at the end - didn't spot it before.

evulins: Done 👍

franiakat (approved):
🥂

PR #2, opened 2021-11-05
Add eslint and prettier

bazelajoanna:
ESlint and prettier has been added based on documentation https://github.com/typescript-eslint/typescript-eslint/blob/master/docs/getting-started/linting/README.md.


.eslintignore
```diff
@@ -0,0 +1,6 @@
+# don't ever lint node_modules
+node_modules
+# don't lint build output (make sure it's set to your correct build folder name)
+dist
+# don't lint nyc coverage output
+coverage
```
franiakat: We don't have either `coverage` or `dist` folder - build is created in `build` directory, so we can change `dist` to `build`.

franiakat (changes requested):
Just change the `.eslintignore` and you are free to go. :)

.eslintignore
bazelajoanna: Ok, done.

.eslintignore
```diff
@@ -0,0 +1,6 @@
+# don't ever lint node_modules
+node_modules
+# don't lint build output (make sure it's set to your correct build folder name)
+build
+# don't lint nyc coverage output
+coverage
```
franiakat: Coverage can be removed, as we don't have the coverage setup (probably referring to test coverage count) ;)

bazelajoanna: Removed :)

franiakat (approved):
Good to go, please use Squash and Merge option. 🎉

PR #3, opened 2021-11-09
Creating style configuration

evulins:
Basic style configuration where we used:
-  https://styled-components.com/
- @types/styled-components

In this tasks we added general style file, breakpoints and assets (fonts, logo, images).


src/App.tsx
```diff
@@ -1,25 +1,25 @@
...
+import { primarySubtitle, darkGradientBackground } from "./styles/colors.js";
+
+const P = styled.p`
+  color: ${primarySubtitle};
+  font-size: 10em;
+`;
+
+const Container = styled.div`
+  background: ${darkGradientBackground};
+`;
```
franiakat: This is just for testing purposes, otherwise would we have a separate directory to keep e.g. styled paragraph, I assume we would?

src/global-styles.js
```diff
@@ -0,0 +1,97 @@
+import { createGlobalStyle } from "styled-components";
```
franiakat: The name that I usually use for the main css style is either `main` or `app`, but if this is some sort of standard connected with `styled components`, then let's use it. I think Style in the name is needed as later this is used as component in `index.tsx`. It would look weird to import App component 😁
I am wondering where to keep this file, I think it could fit better in `styles` folder (unless styled components standard as well).

In general I was taking a look and I think I'd keep a bit different structure, something along those lines:
```
 assets > styles > breakpoints.js & colors.js
 assets > img > icons > all icons here
 assets > img > all background images here
 assets > fonts > all fonts here
```
WDYT? :)

src/index.tsx
```diff
@@ -1,8 +1,8 @@
 /* eslint @typescript-eslint/no-var-requires: "off" */
 import React from "react";
 import ReactDOM from "react-dom";
-import "./index.css";
 import App from "./App";
+import GlobalStyle from "./global-styles";
```
franiakat: 🚀

src/styles/breakpoints.js
```diff
@@ -0,0 +1,22 @@
...
+  base: 0,
+  mobile: "320px",
+  desktopTablet: "600px",
+  desktopL: "900px",
+  desktopXL: "1280px",
+};
+
+export const minDevice = {
+  base: `(min-width: ${size.base})`,
+  mobile: `(min-width: ${size.mobile})`,
```
franiakat: Shouldn't the min be 0? 🤔 
```suggestion
  mobile: `(min-width: ${size.base})`,
```

src/styles/colors.js
```diff
@@ -0,0 +1,61 @@
...
+
+const primaryBackground = "#F7F8FC";
+
+const darkGradientBackground =
+  "linear-gradient(93.51deg, #182E7C 8.9%, #030D67 94.31%)";
+
+const lightGradientBackground =
+  "linear-gradient(93.51deg, #1782DC 8.9%, #6AB7F7 94.31%)";
+
+const activeButton = "#3678E3";
```
franiakat: Is it for background or a font color? 🙃 Maybe nor relevant actually.

src/styles/colors.js
```diff
@@ -0,0 +1,61 @@
...
+
+const mainHeader = "#161C56";
+
+const primarySubtitle = "#124CAB";
+
+const seconderySubtitle = "#F88B0C";
+
+const copyrights = "#DCDCDC";
+
+const textDetails = "#757575";
```
franiakat: I'm not sure whether I like those names, can e.g. `tagBackground` be used in other place then in tag background? Also a question to other names.
Anyway, I am fine also to keep it this way and rename it once we use it in particular styles and it's easier to see the proper name for it. ;)

franiakat (changes requested):
Nice to see it set up, good job you two! See my suggestions and let's find a middle ground 😎

src/App.tsx
evulins: Yes, that is just for testing. In the future it will be in separate component/file.

src/global-styles.js
evulins: Ad 1 I would rather stay with name global-styles - it's not connected with styled-components ;) It says more to me - especially that it's js file. 

Ad 2 I like the new structure - good idea.

src/styles/colors.js
evulins: It is just for background only.

src/styles/colors.js
evulins: For now I would leave like this. We talked with @bazelajoanna and we decided that if during the work we will see that is   illogical we will change it.

src/styles/breakpoints.js
franiakat: How about others, should they also be set as "one down", meaning `(min-width: ${size.mobile})` for tablet and so on? 🤔

src/styles/colors.js
franiakat: 🙆‍♀️

src/styles/breakpoints.js
evulins: I misunderstood you before. Actually, I would change it one more time :) In few articles i found that this a good practice - https://jsramblings.com/how-to-use-media-queries-with-styled-components/ But if you think differently, I will be happy if I get to know new solutions :)

franiakat: I see, I think I need to see it in action to have an opinion, given my zero experience with styled components ;) thanks for the article, it makes sense to keep it this way. :)

franiakat (approved):

PR #4, opened 2021-11-10
Add logo

bazelajoanna:
Add two types of logo - for white and dark background.


src/components/Logo/index.tsx
```diff
@@ -0,0 +1,37 @@
+import React, { FC } from "react";
+import styled from "styled-components";
+import { minDevice } from "../../assets/styles/breakpoints";
+
+import logoBlack from "../../assets/img/logo-black.svg";
+import logoWhite from "../../assets/img/logo-white.svg";
```
franiakat: This is a matter of one's prefferd style, but I can share main:
I use an empty line between imports from outside libraries (whatever we installed in node modules) our own files we've created and style files (when I use css). Then the imports would look like that:
```suggestion
import React, { FC } from "react";
import styled from "styled-components";

import { minDevice } from "../../assets/styles/breakpoints";
import logoBlack from "../../assets/img/logo-black.svg";
import logoWhite from "../../assets/img/logo-white.svg";
```

src/components/Logo/index.tsx
```diff
@@ -0,0 +1,37 @@
...
+  display: flex;
+
+  @media ${minDevice.desktopTablet} {
+    width: 158px;
+  }
+
+  @media ${minDevice.desktopL} {
+    width: 185px;
+  }
+`;
```
franiakat, line 19: Very nice, mobile first approach. 🙆‍♀️

src/components/Logo/index.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+const Img = styled.img`
+  width: 100%;
+`;
+
+interface Props {
+  isBackgroundDark: boolean;
+}
+
+const Logo: FC<Props> = ({ isBackgroundDark }) => {
```
franiakat: Or more descriptive:

```suggestion
interface LogoProps {
  isBackgroundDark: boolean;
}

const Logo: FC<LogoProps> = ({ isBackgroundDark }) => {
```

src/components/P/index.tsx
```diff
@@ -0,0 +1,29 @@
+import React from "react";
+import styled from "styled-components";
+import PropTypes from "prop-types";
```
franiakat:
```suggestion
```

src/components/P/index.tsx
```diff
@@ -0,0 +1,29 @@
...
+  fontSize?: string;
+  fontWeight?: string;
+  color?: string;
+  textAlign?: string;
+  marginLeft?: string;
+  marginBottom?: string;
+  marginTop?: string;
+  width?: string;
+  shadow?: string;
+}
```
franiakat: Interesting fact, in case you need to have an interface where all those properties are required and another one where all are optional you could write it this way:
```
interface Props {
  fontSize: string;
  fontWeight: string;
  color: string;
  textAlign: string;
  marginLeft: string;
  marginBottom: string;
  marginTop: string;
  width: string;
  shadow: string;
}

interface PropsOptional extends Partial<Props>
```

src/components/P/index.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+const P = styled.p<Props>`
```
franiakat: You don't use it anywhere right now, right?

src/components/P/index.tsx
```diff
@@ -0,0 +1,29 @@
+import React from "react";
+import styled from "styled-components";
+import PropTypes from "prop-types";
+
+interface Props {
```
franiakat: Normally I'd make it more descriptive and use the name of the element as well. Here it is a bit tricky as it would be `PProps` 😁  I'd still do it to keep the convention, I think. 🤔

franiakat (commented):
🐎  Very chic first elemnt, nice! Please, see the comments :)

franiakat:
As a side note you can use interactive rebase to clean up in commits a bit, in the end we don't want commits with wip word on master 😎  (Even if they are squashed 🌚 )

src/components/P/index.tsx
franiakat: https://www.typescriptlang.org/docs/handbook/utility-types.html

src/components/P/index.tsx
bazelajoanna: Yes, at first, we thought we could use it within the Logo component (for the text), but then we changed our minds and imported the whole logo.

franiakat (approved):

PR #5, opened 2021-11-14
Add footer

evulins:
Add and style responsive footer.


src/components/Footer/styledComponents.tsx
```diff
@@ -0,0 +1,64 @@
...
+  color: ${white};
+  @media ${minDevice.desktopTablet} {
+    font-size: 1.25em;
+  }
+  @media ${minDevice.desktopL} {
+    font-size: 1.5em;
+  }
+  @media ${minDevice.desktopXL} {
+    font-size: 2em;
+  }
```
bazelajoanna, line 34: Font size for tablet and bigger screen should be 1em (16px) :)

src/components/Footer/index.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+const Footer: FC = () => {
+  return (
+    <FooterWrapper>
+      <Wrapper>
+        <Logo isBackgroundDark width="137px" />
+        <ContactWrapper>
+          <FooterP>Kontakt</FooterP>
+          <Row>
+            <Icon color={white} width="1em">
```
bazelajoanna, line 30: For mobile maybe there is no need to set padding on that element? :)

bazelajoanna (commented):
Just small suggestions. Great job! ;)

src/components/Footer/index.tsx
evulins: Good point :)

src/components/Footer/styledComponents.tsx
evulins: Good point :)

src/App.tsx
```diff
@@ -1,10 +1,10 @@
 import React, { FC } from "react";
-import Logo from "./components/Logo";
+import Footer from "./components/Footer";
 
 const App: FC = () => {
   return (
     <div>
```
franiakat: Is this div needed? : >

src/components/Footer/index.tsx
evulins: Done :)

src/components/Footer/styledComponents.tsx
```diff
@@ -0,0 +1,58 @@
+import styled from "styled-components";
```
franiakat, line 1: I am wondering where should we store these and about the name of this file, is it done the way you are used to? :)
For now this is relevant only for Footer component and we should not optimise for the unknown future, so probably this is just fine 😊

src/components/Footer/styledComponents.tsx
```diff
@@ -0,0 +1,58 @@
...
+  width: 100%;
+  height: auto;
+  display: flex;
+  flex-direction: column;
+  position: absolute;
+  bottom: 0;
+  background: ${darkGradientBackground};
+`;
+
+const Wrapper = styled.div`
```
franiakat: I am thinking if there is the way to have a base wrapper element that we could then use as as base for other variants. I see that sth like column handles the basic redundant styling:
```
const Column = styled.div`
  display: flex;
  flex-direction: column;
`;
```
Do you think it would be then more DRY (Don't Repeat Yourself) but still good to read? Or only DRY? 😁

src/components/Footer/styledComponents.tsx
```diff
@@ -0,0 +1,58 @@
+import styled from "styled-components";
+
+import P from "../P";
+import { minDevice } from "../../assets/styles/breakpoints";
+import { darkGradientBackground, white } from "../../assets/styles/colors";
+
+const FooterWrapper = styled.div`
+  width: 100%;
+  height: auto;
```
franiakat: I am wondering whether this is actually needed, `auto` is the default anyway. 🤔  removing it, does not affect the UI.

src/components/Footer/index.tsx
```diff
@@ -0,0 +1,44 @@
...
+} from "./styledComponents";
+import { white, copyrights } from "../../assets/styles/colors";
+
+const Footer: FC = () => {
+  return (
+    <FooterWrapper>
+      <Wrapper>
+        <Logo isBackgroundDark width="137px" />
+        <ContactWrapper>
+          <FooterP>Kontakt</FooterP>
```
franiakat, line 28: I think there is 4px padding missing between Kontakt and email address.

src/components/Footer/index.tsx
```diff
@@ -0,0 +1,44 @@
...
+            </Icon>
+            <MailA href="mailto:myslaceklasy@email.com">
+              myslaceklasy@email.com
+            </MailA>
+          </Row>
+        </ContactWrapper>
+      </Wrapper>
+      <Column>
+        <P color={copyrights} textAlign="center" fontSize="0.875em">
+          © Copyrights 2020.
```
franiakat, line 46: I think there is 4 px bottom padding missing here, but I think line-height is bigger 28px instead of 24px I think 😀

src/components/Icon/index.tsx
```diff
@@ -0,0 +1,31 @@
...
+  color: string;
+  width: string;
+  hoverColor: string;
+  paddingRight: string;
+}
+
+type PropsOptional = Partial<IconProps>;
+
+const Icon = styled.div<PropsOptional>`
+  color: ${(props) => props.color || primaryText};
```
franiakat, line 10: Meta thought - are you used to have arrow parens always? The thought behind that rule is described here:
```
Always include parentheses around arguments for clarity and consistency. eslint: arrow-parens
Why? Minimizes diff churn when adding or removing arguments.
```
https://github.com/airbnb/javascript#arrows--one-arg-parens

src/components/Logo/index.tsx
```diff
@@ -24,11 +28,12 @@ const Img = styled.img`
 
 interface LogoProps {
   isBackgroundDark: boolean;
+  width?: string;
```
franiakat: Normally I always put interfaces at the beginning of the file, this way you already know what are the needs of given element. We can agree to some convention, maybe we'd like to extract interfaces to separate files/keep them grouped at the top or leave it this way? @evulins @bazelajoanna

src/components/Row/index.tsx
```diff
@@ -0,0 +1,8 @@
+import styled from "styled-components";
+
+const Row = styled.div`
+  display: flex;
+  flex-direction: row;
+`;
+
+export default Row;
```
franiakat, line 8: 👍

franiakat (commented):
🏇 👌

src/App.tsx
evulins: It just leftovers from the main branch. It can be easily deleted.

src/components/Footer/styledComponents.tsx
evulins: Those are just components for Footer so IMO it should be stored in Footer folder. I am not attached to the name of the file - I could find better one for the file where we keep styled components ;)

src/components/Footer/index.tsx
evulins: I added the padding-bottom but I left line-height as it was - 24px. I think line-height here depends on where you check in on the design ;) In this view on every device is 24px: https://www.figma.com/file/FZKtRjUFs3QHXZgQKnJRMW/Thinking-Classroom?node-id=474%3A187

src/components/Footer/index.tsx
evulins: Done

src/components/Footer/styledComponents.tsx
evulins: I can prepare some base wrapper and we can modify it during adding more components.

src/components/Logo/index.tsx
evulins: I am ok with both - moving to the top and also moving to a different file.

src/components/Footer/styledComponents.tsx
franiakat: We will agree on this convention later on, if that's fine ;)

src/components/Logo/index.tsx
franiakat: FYI Agreed to extract it.

src/components/Footer/index.tsx
franiakat: Oh no, misunderstanding :D I meant it should be 24 px, but I thought it was 28px in your implementation, nvm, all fine now.

src/components/Wrapper/index.tsx
```diff
@@ -0,0 +1,27 @@
+import styled from "styled-components";
+import { minDevice } from "../../assets/styles/breakpoints";
+
+interface WraperProps {
+  flexDirection: string;
+  flexDirectionDT: string;
```
franiakat, line 6: Why DT? : >

evulins: for DesktopTablet breakpoint :D

franiakat: Alright :D

franiakat (approved):
Awesome 😎  Happy to have the footer already, good job!
sqash

bazelajoanna (approved):
:)

PR #6, opened 2021-11-15
Add navbar

bazelajoanna:
Add navbar with hamburger menu WIP


src/components/Navbar/index.tsx
```diff
@@ -0,0 +1,115 @@
...
+    }
+  }
+`;
+
+const HamburgerImg = styled.img`
+  width: 0.98em;
+  @media ${minDevice.desktopTablet} {
+    display: none;
+  }
+`;
```
evulins: Maybe it would be good to move all styled components to a new file. Then we will have a cleaner Navbar component only with functionality. What do you think?

src/components/Navbar/index.tsx
```diff
@@ -0,0 +1,115 @@
...
+  const toggleHamburger = () => {
+    if (hamburgerClass === "") {
+      setHamburgerClass("active");
+      setMenuClass("active");
+      setNavbarClass("active");
+    } else {
+      setHamburgerClass("");
+      setMenuClass("");
+    }
+  };
```
evulins, line 27: Looks good! :)

evulins (commented):
Great job 👍 I have just two suggestions :)

src/components/Navbar/index.tsx
```diff
@@ -0,0 +1,115 @@
...
+
+  @media ${minDevice.desktopTablet} {
+    flex-direction: row;
+    transform: translateY(-0%);
+    position: relative;
+    align-items: center;
+  }
+`;
+
+const NavbarMenuItem = styled(P)`
```
evulins: Styled hoover is missing on the menu

bazelajoanna: Yes, thanks. I will add it :)

src/components/Navbar/index.tsx
bazelajoanna: Sure, thanks. I will create Styled components file in the component folder.

evulins (approved):

PR #7, opened 2021-11-16
Add navbar

bazelajoanna:
Add navbar with hamburger menu

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -11,6 +11,9 @@ module.exports = {
     "plugin:react/recommended",
     "plugin:import/typescript",
   ],
+  rules: {
+    "no-debugger": "off",
+  },
   env: {
     browser: true,
     es6: true,
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,11 +1,12 @@
 import React, { FC } from "react";
+
 import Footer from "./components/Footer";
 import Navbar from "./components/Navbar";
 
 const App: FC = () => {
   return (
     <>
-      <Navbar></Navbar>
+      <Navbar />
       <Footer />
     </>
   );
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,8 +1,14 @@
 import React, { FC } from "react";
 import Footer from "./components/Footer";
+import Navbar from "./components/Navbar";
 
 const App: FC = () => {
-  return <Footer />;
+  return (
+    <>
+      <Navbar></Navbar>
+      <Footer />
+    </>
+  );
 };
 
 export default App;
--- a/src/components/Navbar/styledComponents.tsx
+++ b/src/components/Navbar/styledComponents.tsx
@@ -1,7 +1,7 @@
 import styled from "styled-components";
+
 import { minDevice } from "../../assets/styles/breakpoints";
 import { activeLink } from "../../assets/styles/colors";
-import P from "../P";
 
 const NavbarDiv = styled.div`
   height: 5em;
@@ -29,7 +29,17 @@ const NavbarDiv = styled.div`
     margin: 0 12.31em 0 12.5em;
   }
 `;
-const NavbarMenu = styled.div`
+const NavbarMenu = styled.nav`
+  transform: translateY(-150%);
+  position: absolute;
+
+  @media ${minDevice.desktopTablet} {
+    transform: translateY(-0%);
+    position: relative;
+  }
+`;
+
+const NavbarMenuList = styled.ol`
   display: flex;
   flex-direction: column;
   justify-content: space-evenly;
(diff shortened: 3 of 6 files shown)


evulins (approved):

.eslintrc.js
```diff
@@ -11,6 +11,9 @@ module.exports = {
     "plugin:react/recommended",
     "plugin:import/typescript",
   ],
+  rules: {
+    "no-debugger": "off",
```
franiakat, line 15: <3

src/App.tsx
```diff
@@ -1,8 +1,14 @@
 import React, { FC } from "react";
 import Footer from "./components/Footer";
+import Navbar from "./components/Navbar";
 
 const App: FC = () => {
-  return <Footer />;
+  return (
+    <>
+      <Navbar></Navbar>
```
franiakat:
```suggestion
      <Navbar />
```

src/App.tsx
```diff
@@ -1,8 +1,14 @@
 import React, { FC } from "react";
 import Footer from "./components/Footer";
```
franiakat:
```suggestion
import React, { FC } from "react";

import Footer from "./components/Footer";
```

src/components/Navbar/styledComponents.tsx
```diff
@@ -0,0 +1,98 @@
+import styled from "styled-components";
+import { minDevice } from "../../assets/styles/breakpoints";
```
franiakat:
```suggestion
import styled from "styled-components";

import { minDevice } from "../../assets/styles/breakpoints";
```

src/components/Navbar/styledComponents.tsx
```diff
@@ -0,0 +1,98 @@
...
+import { activeLink } from "../../assets/styles/colors";
+import P from "../P";
+
+const NavbarDiv = styled.div`
+  height: 5em;
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  padding: 0 1.16em 0 0.81em;
```
franiakat, line 11: Once we have more values we may think about extracting some regularly used padding values to contants. ;)

src/components/Navbar/styledComponents.tsx
```diff
@@ -0,0 +1,98 @@
...
+    height: 4.5em;
+  }
+
+  @media ${minDevice.desktopXL} {
+    margin: 0 12.31em 0 12.5em;
+  }
+`;
+const NavbarMenu = styled.div`
+  display: flex;
+  flex-direction: column;
```
franiakat, line 44: I think @evulins extracted a Column recently, do you think this would be useful, or rather a noise here?

src/components/Navbar/styledComponents.tsx
```diff
@@ -0,0 +1,98 @@
...
+  @media ${minDevice.desktopL} {
+    margin: 0 6.25em 0 6.06em;
+    height: 4.5em;
+  }
+
+  @media ${minDevice.desktopXL} {
+    margin: 0 12.31em 0 12.5em;
+  }
+`;
+const NavbarMenu = styled.div`
```
franiakat: We can improve this navigation when it comes to html semantics. ;)
You can taka look at this piece of docs: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/nav
I'd go with `<nav> <ol> and <li>`, you don't need links for now, as we'll use Router. ;)

franiakat (changes requested):
Good job, Asia! 👩‍💻 Take a look at my suggestions ;)
And as discussed on slack, I refactored the Nabar a bit. ;)

franiakat:
Oh and very neat commits history here! 😎

src/components/Navbar/styledComponents.tsx
bazelajoanna: Ok, we can discuss it ;)

src/components/Navbar/styledComponents.tsx
franiakat: Also a tiny read around it: https://dev.to/ameseee/semantic-html-how-can-you-tell-13am

bazelajoanna: Thanks, I've made these changes :)

src/components/Navbar/styledComponents.tsx
bazelajoanna: Hm, I don't know. Maybe here there is no need to use this component?

franiakat: Yeah, it is fine not to, now that I think of it, I wouldn't either 👯‍♂️

src/components/Navbar/styledComponents.tsx
```diff
@@ -0,0 +1,110 @@
+import styled from "styled-components";
```
franiakat, line 1: Very nice implementation 💪

franiakat (approved):
Looks cool!
sqash

PR #11, opened 2021-11-17
Add Result Count Label

bazelajoanna:
Add result count label component.


src/App.tsx
```diff
@@ -2,11 +2,13 @@ import React, { FC } from "react";
```
franiakat, line 2: React import is not needed anymore. : >

src/components/ResultCountLabel/index.tsx
```diff
@@ -0,0 +1,20 @@
+import React, { FC } from "react";
```
franiakat: React import is not needed anymore. : >

src/components/ResultCountLabel/index.tsx
```diff
@@ -0,0 +1,20 @@
...
+import styled from "styled-components";
```
franiakat: `styled` is not used here, can be removed.

src/components/ResultCountLabel/index.tsx
```diff
@@ -0,0 +1,20 @@
...
+
+import { ResultCountWrapper, ResultCountP } from "./styledComponents";
+import { ResultCountWrapperProps } from "./interface";
+
+const ResultCountLabel: FC<ResultCountWrapperProps> = ({ count }) => {
+  return (
+    <ResultCountWrapper
```
franiakat: Do you think those values has to be customisable from the outside? I wonder whether there is ever the case that flex direction or any other prop would be defined differently than right here? If never, then i assume that could be just written in styles, right? :)

src/components/ResultCountLabel/interface.ts
```diff
@@ -0,0 +1,3 @@
+export interface ResultCountWrapperProps {
+  count: number;
+}
```
franiakat, line 3: 🥳

src/components/ResultCountLabel/styledComponents.tsx
```diff
@@ -0,0 +1,45 @@
+import styled from "styled-components";
```
franiakat, line 1: Looks nice! :)

franiakat (changes requested):
Good job, looks legit! :)
I had a few tiny comments to consider.

franiakat:
BTW You can consider cleaning up history a bit (WIP :D)

franiakat:
P.P.S Please add a test that checks whether the passed count is being displayed - let's pair to write it together once you are done with code review changes. 😊

evulins (changes requested):
Great job! Looks nice on a variety of devices. With tests, it will be a front-end masterpiece :)

src/components/ResultCountLabel/index.tsx
franiakat: I mean all props, flexDirection, justifyContent, alignItems, padding? Is it fine to just permanently define them in the ResultCountWrapper styling, instead of passing them from the outside? :> We can also tak it through over the call.

bazelajoanna: I've moved those values to styledComponents file :)

franiakat (approved):
sqash

PR #12, opened 2021-11-17
Add tag component

bazelajoanna:
Add tag component.


src/components/Tag/index.tsx
```diff
@@ -0,0 +1,19 @@
...
+import { TagWrapper, SubtractComponent } from "./styledComponents";
+import { TagNameProps } from "./interface";
+
+const Tag: FC<TagNameProps> = ({ text, removeTag }) => {
+  return (
+    <TagWrapper
+      flexDirection="row"
+      justifyContent="space-between"
+      alignItems="center"
+    >
```
franiakat: We don't need to have it customisable from the outside - it never changes in design, please put it in styles.

src/components/Tag/interface.ts
```diff
@@ -0,0 +1,4 @@
+export interface TagNameProps {
```
franiakat:
```suggestion
export interface TagProps {
```

src/components/Tag/styledComponents.tsx
```diff
@@ -0,0 +1,27 @@
...
+  width: fit-content;
+  background-color: ${tagBackground};
+  border-radius: 0.938em;
+  line-height: 1.5em;
+  font-size: 1em;
+  color: ${tagText};
+  font-weight: 700;
+`;
+
+const SubtractComponent = styled(Subtract)`
```
franiakat:
```suggestion
const RemoveTagIcon = styled(RemoveIcon)`
```

franiakat (approved):
sqash

PR #13, opened 2021-11-18
Add input with problem search

evulins:
Input with autocompletes suggestions component.


src/App.tsx
```diff
@@ -4,14 +4,17 @@ import Footer from "./components/Footer";
 import Navbar from "./components/Navbar";
 import ResultCountLabel from "./components/ResultCountLabel";
 import Tag from "./components/Tag";
+import SearchingSection from "./components/SearchingSection";
 
 const App: FC = () => {
   return (
     <>
       <Navbar />
+      <SearchingSection />
```
franiakat: Or more descriptive? A suggestion:
```suggestion
      <ProblemSearchSection />
```

src/components/Input/index.tsx
```diff
@@ -0,0 +1,121 @@
...
+          {showClearButton && searchTerm && (
+            <Button aria-label="Remove button" onClick={() => clearInput()}>
+              <span id="Remove button" hidden>
+                Remove input text button
+              </span>
+
+              <RemoveIcon aria-hidden="true" focusable="false" />
+            </Button>
+          )}
+        </Label>
```
franiakat: We'd rather want to have a label next to the input and connect them with `htmlFor` property on label and `id` on input:
```
<Label htmlFor="input-search" id="input-search-label">
    <SearchIcon />
</Label>
<Input
     id="input-search"
/>
```

src/components/Input/index.tsx
```diff
@@ -0,0 +1,121 @@
...
+        method="get"
+        onSubmit={(e) => {
+          e.preventDefault();
+        }}
+      >
+        <Label htmlFor="header-search">
+          <Input
+            type="text"
+            placeholder="Szukaj problemów matematycznych"
+            aria-label="search-input"
```
franiakat: It should be more human-friendly ;)
```suggestion
            aria-label="Szukaj problemów"
```

src/components/Input/index.tsx
```diff
@@ -0,0 +1,121 @@
...
+    } else if (e.key === "ArrowDown") {
+      if (activeSuggestionIndex < filteredSuggestions.length) {
+        e.preventDefault();
+        setActiveSuggestionIndex(activeSuggestionIndex + 1);
+      }
+    }
+  };
+
+  return (
+    <FormWrapper>
```
franiakat: We need to restructure an html a bit to comply to ARIA and HTML standards:
https://www.w3.org/TR/wai-aria-practices/examples/combobox/aria1.1pattern/listbox-combo.html
We could have a div named e.g. `Combobox` and its properties are as follows:
```
<Combobox
      id="combobox"
      aria-expanded="false"
      aria-owns="autocomplete-options"
      aria-haspopup="listbox"
>    
```

src/components/Input/index.tsx
```diff
@@ -0,0 +1,121 @@
...
+            placeholder="Szukaj problemów matematycznych"
+            aria-label="search-input"
+            aria-owns="autocomplete-options"
+            aria-autocomplete="list"
+            value={searchTerm}
+            onChange={handleChange}
+            onKeyDown={(e) => onKeyDown(e)}
+            onSubmit={(e) => {
+              e.preventDefault();
+            }}
```
franiakat:
```suggestion
          <Input
            id="input-search"
            type="text"
            placeholder="Szukaj problemów matematycznych"
            aria-label="search-input"
            aria-autocomplete="list"
            aria-controls="autocomplete-options"
            value={searchTerm}
            onChange={handleChange}
            onKeyDown={(e) => onKeyDown(e)}
            onSubmit={(e) => {
              e.preventDefault();
            }}
```

src/components/Input/suggestionsListComponent.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+const SuggestionsListComponent: FC<SuggestionsProps> = ({
+  filteredSuggestions,
+  activeSuggestionIndex,
+  chooseSuggestion,
+}) => {
+  const ref = useRef<any>(); //I didn't find proper type for this one.
+
+  return filteredSuggestions.length ? (
+    <SuggestionList ref={ref}>
```
franiakat:
```suggestion
    <SuggestionList 
         ref={ref}
         id="autocomplete-options"
         aria-labelledby="search-input-label"
         role="listbox"
    >
```

src/components/Input/suggestionsListComponent.tsx
```diff
@@ -0,0 +1,50 @@
...
+            key={suggestion}
+            onClick={(e: React.MouseEvent<HTMLLIElement>) =>
+              chooseSuggestion(e)
+            }
+          >
+            <Clipboard style={{ paddingRight: "0.5em" }} />
+            <span data-test-id="suggestion-text">{suggestion}</span>
+          </Suggestion>
+        );
+      })}
```
franiakat:
```suggestion
      {filteredSuggestions.map((suggestion: string, index: number) => (
            <Suggestion
              key={suggestion}
              onClick={(e: React.MouseEvent<HTMLLIElement>) =>
                chooseSuggestion(e)
              }
              isSelected={index === activeSuggestionIndex}
            >
              <Clipboard style={{ paddingRight: "0.5em" }} />
              <span data-test-id="suggestion-text">{suggestion}</span>
            </Suggestion>
      ))}
```

src/components/Input/index.tsx
```diff
@@ -0,0 +1,121 @@
...
+            />
+          )}
+          <SearchIcon />
+          {showClearButton && searchTerm && (
+            <Button aria-label="Remove button" onClick={() => clearInput()}>
+              <span id="Remove button" hidden>
+                Remove input text button
+              </span>
+
+              <RemoveIcon aria-hidden="true" focusable="false" />
```
franiakat: Hmm, I think we should remove those properties suggesting the element should not be interactive, as you should be able to interact with it using keyboard. ;)
https://developer.mozilla.org/en-US/docs/Web/Accessibility/Understanding_WCAG/Keyboard

franiakat (changes requested):
TBC :D

src/components/Input/index.tsx
```diff
@@ -0,0 +1,121 @@
...
+  Form,
+  Input,
+  Label,
+  SearchIcon,
+  RemoveIcon,
+  FormWrapper,
+  Button,
+} from "./styledComponents";
+
+const SearchInput: FC<SearchInputProps> = ({ suggestions }) => {
```
franiakat: I'd rename the component and the file to something more descriptive of this functionality, like: `SearchAutocomplete`, `ListAutocomplete`, `Combobox`? WDYT?

src/components/Input/interface.ts
```diff
@@ -0,0 +1,13 @@
+export interface SearchInputProps {
+  suggestions: Array<string>;
```
franiakat: Those are more like `items` in a sense: Problems and Tags, only after interacting with the input search you filter them and then the limited group of them become `suggestions` to your search term.

src/components/Input/interface.ts
```diff
@@ -0,0 +1,13 @@
...
+}
+
+export interface SuggestionsProps {
+  filteredSuggestions: Array<string>;
```
franiakat: So then those would be `suggestions` or `filteredItems`?

src/components/Input/interface.ts
```diff
@@ -0,0 +1,13 @@
...
+  activeSuggestionIndex: number;
+  chooseSuggestion: (e: React.MouseEvent<HTMLLIElement, MouseEvent>) => void;
```
franiakat: But those are still suggestions, I think - we can talk this through on the call.

src/components/Input/interface.ts
```diff
@@ -0,0 +1,13 @@
...
+}
+
+export interface SuggestionProps {
+  isSelected?: boolean;
```
franiakat: This is always present now.

src/components/Input/styledComponents.tsx
```diff
@@ -0,0 +1,118 @@
+import styled, { css } from "styled-components";
+
+import { SuggestionProps } from "./interface";
+
```
franiakat:
```suggestion
```

franiakat: Unless we're adding a new line before assets import.

src/components/Input/styledComponents.tsx
```diff
@@ -0,0 +1,118 @@
...
+const FormWrapper = styled.div`
+  width: 100%;
+  height: 5em;
+  box-shadow: 0px 4px 15px 0px rgba(22, 28, 86, 0.2);
+  z-index: 10;
+  @media ${minDevice.desktopTablet} {
+    width: 22.5em;
+  }
+  @media ${minDevice.desktopL} {
+    width: 31.4375em;
```
franiakat: We can talk about using px, em, rems in particular use cases and come up with some nice `rules` that could help us reflecting our design in a convenient way :)

src/components/Input/index.tsx
evulins: I could go with SearchAutocomplete or Combobox

src/components/Input/interface.ts
evulins: Yes, let's talk about it :)

src/components/SearchingSection/index.tsx
```diff
@@ -0,0 +1,44 @@
...
+    >
+      <P
+        color={white}
+        paddingBottom="1em"
+        fontSize="1.125em"
+        fontSizeDL="1.5em"
+        fontWeight="700"
+      >
+        Szukaj problemów matematycznych
+      </P>
```
franiakat: I see it more as a heading not a paragraph.

src/components/Input/styledComponents.tsx
evulins: This was my way of thinking to keep those things seperated.

src/components/Input/index.test.tsx
```diff
@@ -0,0 +1,189 @@
...
+  screen,
+  fireEvent,
+  within,
+  cleanup,
+} from "@testing-library/react";
+import Input from "./index";
+import { Suggestion, SuggestionList } from "./styledComponents";
+
+describe("Input component", () => {
+  it("should allow letters to be inputted", () => {
```
franiakat: or `displays inputted value`?

src/components/Input/index.test.tsx
```diff
@@ -0,0 +1,189 @@
...
+      />
+    );
+
+    let input = screen.getByLabelText("search-input") as HTMLInputElement;
+    expect(input.value).toBe("");
+    fireEvent.change(input, { target: { value: "Good Day" } });
+    expect(input.value).toBe("Good Day");
+  });
+
+  it("should show suggestions list matching input target", () => {
```
franiakat:
```suggestion
  it("should show suggestions list matching input value", () => {
```

src/components/Input/index.test.tsx
```diff
@@ -0,0 +1,189 @@
...
+    fireEvent.change(input, { target: { value: "e" } });
+    expect(input.value).toBe("e");
+
+    const list = screen.getByRole("list");
+    const { getAllByRole } = within(list);
+    const items = getAllByRole("listitem");
+    expect(items.length).toBe(4);
+  });
+
+  it("should show in input field chosen suggestion", () => {
```
franiakat:
```suggestion
  it("should display picked suggestion as input value", () => {
```

src/components/Input/index.test.tsx
```diff
@@ -0,0 +1,189 @@
...
+    let input = screen.getByLabelText("search-input") as HTMLInputElement;
+
+    expect(input.value).toBe("");
+    fireEvent.change(input, { target: { value: "e" } });
+    expect(input.value).toBe("e");
+
+    const list = screen.getByRole("list");
+    const { getAllByRole } = within(list);
+    const items = getAllByRole("listitem");
+    expect(items.length).toBe(4);
```
bazelajoanna: We can also check the `items` length before typing `e` to test if there is no items.

src/components/Input/index.test.tsx
```diff
@@ -0,0 +1,189 @@
...
+
+    const list = screen.getByRole("list");
+    const { getAllByRole } = within(list);
+    const items = getAllByRole("listitem");
+
+    expect(items.length).toBe(1);
+
+    const item = screen.getByText("Alligator").textContent;
+
+    fireEvent.change(input, { target: { value: item } });
```
bazelajoanna: We'd like also to check if selected item displays in input after clicking on it.

evulins: I would need help with it cause  I was thinking about it before and the only way i test it is with checking input value after all

src/components/Input/index.test.tsx
evulins: I can't check the items length and list because they are not in the document.

src/components/Input/index.test.tsx
```diff
@@ -0,0 +1,189 @@
...
+    expect(input.value).toBe("Alligator");
+  });
+
+  it("shouldn't show remove button in input field", () => {
```
bazelajoanna: These two tests (from line 100 and 125) can be merged into one. Maybe named as "it shows remove button after suggestion list is displayed"?

src/components/SearchAutocomplete/index.tsx
```diff
@@ -0,0 +1,127 @@
...
+  return (
+    <Combobox
+      id="combobox"
+      aria-expanded="false"
+      aria-owns="autocomplete-options"
+      aria-haspopup="listbox"
+    >
+      <Form
+        action="/"
+        method="get"
```
franiakat: This can be removed, we're not really "submitting" this form anywhere and `action` should point to a resource that would process the form data after submission, `method` is not needed as well for the same reason, btw. get is the default one. ;)

src/components/SearchAutocomplete/index.tsx
```diff
@@ -0,0 +1,127 @@
...
+    setFilteredSuggestions([]);
+    setSearchTerm(e.currentTarget.innerText);
+    setActiveSuggestionIndex(0);
+    setShowSuggestions(false);
+  };
+
+  //It's working, but I would like to prevent behaviour that
+  //after I hit the last suggestion the background style don't vanish
+
+  const onKeyDown = (e: KeyboardEvent) => {
```
franiakat:
```suggestion
  const handleKeyDown = (e: KeyboardEvent) => {
```

src/components/SearchAutocomplete/index.tsx
```diff
@@ -0,0 +1,127 @@
...
+          {showClearButton && searchTerm && (
+            <Button aria-label="Remove button" onClick={() => clearInput()}>
+              <span id="Remove button" hidden>
+                Remove input text button
+              </span>
+
+              <RemoveIcon aria-hidden="true" />
+            </Button>
+          )}
+        </Label>
```
franiakat: I wouldn't wrap the clear button into the label, those serve two different purposes, so I'd have a label with serach icon and a clear button as a sibling as well. ;)

src/components/Input/index.test.tsx
franiakat: Right, you could then expect there is no list in the document ;)

src/components/Input/index.test.tsx
franiakat: I was expecting that, that's never straight-forward 😅  let's try to deal with it tomorrow 💪

src/components/SearchAutocomplete/index.tsx
```diff
@@ -0,0 +1,127 @@
...
+  const chooseSuggestion = (e: React.MouseEvent<HTMLLIElement, MouseEvent>) => {
+    e.preventDefault();
+    setFilteredSuggestions([]);
+    setSearchTerm(e.currentTarget.innerText);
+    setActiveSuggestionIndex(0);
+    setShowSuggestions(false);
+  };
+
+  //It's working, but I would like to prevent behaviour that
+  //after I hit the last suggestion the background style don't vanish
```
franiakat: Let's check that tomorrow as well.

franiakat (approved):

PR #15, opened 2021-11-24
Add buttons

bazelajoanna:
Create Button component that handles all the management around disabled state, primary and non primary styling, download icon rendering and children rendering.


bazelajoanna:
@evulins - ready to be reviewed :)

src/assets/styles/colors.js
```diff
@@ -32,7 +34,7 @@ const mainHeader = "#161C56";
 
 const primarySubtitle = "#124CAB";
 
-const seconderySubtitle = "#F88B0C";
+const secondarySubtitle = "#F88B0C";
```
evulins, line 37: 👍

src/components/Button/interface.tsx
```diff
@@ -0,0 +1,5 @@
+export interface ButtonProps {
```
evulins: Could we use in this case` type PropsOptional`?

src/components/Button/styledComponents.tsx
```diff
@@ -0,0 +1,87 @@
...
+import {
+  activeButton,
+  hoverButton,
+  disabledButton,
+  hoverSecondaryButton,
+} from "../../assets/styles/colors";
+import { ReactComponent as Download } from "../../assets/img/icons/download.svg";
+import { minDevice } from "../../assets/styles/breakpoints";
+
+interface ButtonStyleProps {
```
evulins: How about moving it to the interface file? Is this making any sense?

src/components/Button/styledComponents.tsx
```diff
@@ -0,0 +1,87 @@
...
+    background-color: ${disabledButton};
+    box-shadow: none;
+  }
+
+  @media ${minDevice.desktopL} {
+    padding: 16px 48px;
+    font-size: 1rem;
+  }
+
+  ${(props) =>
```
evulins, line 47: 🥇

src/components/ProblemIndex/index.tsx
```diff
@@ -0,0 +1,29 @@
+import { FC } from "react";
+
+import ResultCountLabel from "../ResultCountLabel";
+import Tag from "../Tag";
+import Button from "../Button";
+
+const ProblemIndex: FC = () => {
```
evulins, line 7: Great idea!

src/components/Wrapper/interface.ts
```diff
@@ -0,0 +1,12 @@
+interface WraperProps {
```
evulins, line 1: 👍

evulins (changes requested):
Great job! I have only few questions :)

src/components/Button/interface.tsx
bazelajoanna: We don't need any of these props to be required elsewhere and having them optional is the only version of interface we need in our app - that's why we don't need to create another type with using PropsOptional.

src/components/Button/styledComponents.tsx
bazelajoanna: Yes, good point :) I've moved it there.

evulins (approved):

PR #17, opened 2021-12-01
Refactor styled components

evulins:
Refactor structure in the project, names of the files, and way of importing styled components to match those rules:

- we use atomic design: atoms, molecules, organisms (possibly templates) and pages
- import * as S from ‘./styles’
- styling files are named: styles.ts
- single component directory structure is:
        -  components/Button/index.ts
        -  components/Button/styles.ts
- for components only with styling:
         -  components/Button/index.ts


franiakat:
Aweesome! 🎉 I like te description as well 😻 
A few comments:
- I think Navbar and Footer  and possibly ProblemSearchSection are organisms
- Problem index is rather a page component

Here is quite nice article around atomic design (edit: not true - from its creator) https://tsh.io/blog/atomic-design-explained-by-ryan-reynolds/

src/App.tsx
```diff
@@ -1,9 +1,9 @@
 import { FC } from "react";
 
 import Navbar from "./components/atoms/Navbar";
-import Footer from "./components/molecules/Footer";
-import ProblemIndex from "./components/organisms/ProblemIndex";
-import ProblemSearchSection from "./components/molecules/ProblemSearchSection";
+import Footer from "./components/organisms/Footer";
+import ProblemIndex from "./components/ProblemIndex";
```
franiakat: We can have components/pages/ProblemIndex.

franiakat:
Once the pages change is done, this is good to go :)

franiakat (approved):
🌮

PR #19, opened 2021-12-03
Fix footer position and Input radius

evulins:
Fix styles in Footer and Input components.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -2,7 +2,7 @@ import { FC } from "react";
 
 import Navbar from "./components/atoms/Navbar";
 import Footer from "./components/organisms/Footer";
-import ProblemIndex from "./components/pages/ProblemIndex";
+import ProblemIndex from "./components/ProblemIndex";
 import ProblemSearchSection from "./components/organisms/ProblemSearchSection";
 
 const App: FC = () => {
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -2,7 +2,7 @@ import { FC } from "react";
 
 import Navbar from "./components/atoms/Navbar";
 import Footer from "./components/organisms/Footer";
-import ProblemIndex from "./components/ProblemIndex";
+import ProblemIndex from "./components/pages/ProblemIndex";
 import ProblemSearchSection from "./components/organisms/ProblemSearchSection";
 
 const App: FC = () => {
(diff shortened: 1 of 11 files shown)


src/App.tsx
```diff
@@ -2,7 +2,7 @@ import { FC } from "react";
 
 import Navbar from "./components/atoms/Navbar";
 import Footer from "./components/organisms/Footer";
-import ProblemIndex from "./components/pages/ProblemIndex";
+import ProblemIndex from "./components/ProblemIndex";
```
franiakat: We want to keep `pages` within `components`. ;)

src/components/molecules/SearchAutocomplete/index.tsx
```diff
@@ -69,6 +70,7 @@ const SearchAutocomplete: FC<SearchAutocompleteProps> = ({ suggestions }) => {
     >
       <S.Form
         // autoComplete="off"
```
franiakat: I think this isn't needed in the end, we can remove it here as it's tiny. :)

bazelajoanna:
🥇 great!

src/components/molecules/SearchAutocomplete/interface.ts
```diff
@@ -11,3 +11,7 @@ export interface SuggestionsProps {
 export interface SuggestionProps {
   isSelected?: boolean;
 }
+
+export interface InputProps {
+  showSuggestions: boolean;
```
franiakat: I am wondering whether to indicate those are Style related props, like we did in Button case.
```suggestion
export interface InputStyleProps {
  showSuggestions: boolean;
```

franiakat (commented):
Looks marvellous 🌟 Just those tiny question marks to cover. : >

src/App.tsx
evulins: For sure we want this, i don't know what happened here

src/components/molecules/SearchAutocomplete/index.tsx
```diff
@@ -69,8 +69,6 @@ const SearchAutocomplete: FC<SearchAutocompleteProps> = ({ suggestions }) => {
       aria-haspopup="listbox"
     >
       <S.Form
-        // autoComplete="off"
-        showSuggestions={showSuggestions}
```
franiakat: This one stays I guess. 🙃

evulins: I don't think so :D

evulins: Hit refresh

franiakat: ahaha this is a Form component :D NVM

evulins: I forgot to delete it earlier

franiakat (approved):
sqash

PR #20, opened 2021-12-06
Fix hamburger menu

bazelajoanna:
Fix menu transition and appearance.


src/components/atoms/Navbar/index.tsx
```diff
@@ -33,24 +33,26 @@ const Navbar: FC = () => {
...
-          className={navbarMenuStyle}
-          onClick={toggleHamburger}
-          src={isNavbarMenuActive ? hamburgerX : hamburger}
-          aria-hidden
-          alt="Ikona menu"
-        ></S.HamburgerImg>
-      </S.MenuIconButton>
-      <S.NavbarMenu>
-        <S.NavbarMenuList className={navbarMenuStyle}>
+    <S.HeaderWrapper>
```
franiakat, line 36: I think this is an `organism` already.

src/components/atoms/Navbar/styles.tsx
```diff
@@ -1,76 +1,102 @@
...
-import { minDevice } from "../../../assets/styles/breakpoints";
-import { activeLink } from "../../../assets/styles/colors";
 import Wrapper from "../Wrapper";
 
-const NavbarDiv = styled(Wrapper)`
-  height: 8rem;
+import { minDevice } from "../../../assets/styles/breakpoints";
+import { activeLink, white } from "../../../assets/styles/colors";
+
+const HeaderWrapper = styled.div`
```
franiakat: I think we should have `header` tag here instead.

src/components/atoms/Navbar/styles.tsx
```diff
@@ -1,76 +1,102 @@
...
+  position: relative;
+  z-index: 11;
+  display: flex;
+`;
+
+const NavbarWrapper = styled(Wrapper)`
+  height: 8rem;
+  width: 100%;
+  position: relative;
+  z-index: 12;
```
franiakat, line 20: I think this isn't needed.

src/components/atoms/Navbar/styles.tsx
```diff
@@ -1,76 +1,102 @@
...
+  }
+`;
+
+const NavbarMenuList = styled.ul`
+  display: flex;
+  flex-direction: column;
+  justify-content: space-evenly;
+  align-items: flex-start;
+  padding-left: 2rem;
+
```
evulins, line 83: We could add it here width: 100vw to prevent strange behavior while it closing

franiakat (commented):
Tiny things in comments, awesome cleanup and fix 🧚

src/components/atoms/Wrapper/index.tsx
```diff
@@ -1,15 +1,15 @@
...
 
-const Wrapper = styled.div<PropsOptional>`
+const Wrapper = styled.div<WrapperProps>`
   display: flex;
   flex-direction: ${(props) => props.flexDirection || "row"};
   justify-content: ${(props) => props.justifyContent || "inherit"};
   align-items: ${(props) => props.alignItems || "inherit"};
-  padding: ${(props) => props.padding || "1.6rem"};
-  margin: ${(props) => props.margin || "1.6rem"};
+  padding: ${(props) => props.padding || "0 1.6rem"};
```
evulins, line 11: Have you checked if this change in padding doesn't mess with styles in other places?

evulins (commented):
Great job! Looks great! Some small questions :)

src/components/atoms/Navbar/styles.tsx
bazelajoanna: I've removed height and position, but width and z-index are necessary here.

src/components/atoms/Navbar/styles.tsx
bazelajoanna: Good suggestion, thanks! :)

evulins: But only for mobile devices with hamburger :D

franiakat: 🍔

src/components/atoms/Navbar/styles.tsx
franiakat: Can we have a quick call? I've removed it locally and was working well for me, maybe I missed something ;)

src/components/atoms/Navbar/styles.tsx
bazelajoanna: Yes, I've noticed that 👩‍🎓

src/components/atoms/Wrapper/index.tsx
bazelajoanna: Yes, everything works :) I've just changed the line height in FooterP and now it looks even better 🌷

evulins: Wohooooo 🍾

evulins (changes requested):
👍

evulins (approved):

franiakat (approved):
sqash

PR #21, opened 2021-12-07
Setup mock server with Apollo Graphql client

franiakat:
Setup Apollo GraphQL client and use mocked data responses.
Provide possibility to filter problems by tag.


src/apolloClient.ts
```diff
@@ -0,0 +1,18 @@
+import { ApolloClient, HttpLink, InMemoryCache } from "@apollo/client";
+
+const cache = new InMemoryCache();
+
+const link = new HttpLink({
+  uri: "http://localhost:9779",
+
+  // Use explicit `window.fetch` so tha outgoing requests
```
evulins: Do we want to leave those explanations here?

src/assets/img/icons/search.svg
```diff
@@ -1,4 +1,5 @@
 <svg width="19" height="19" viewBox="0 0 19 19" fill="none" xmlns="http://www.w3.org/2000/svg">
 <path d="M8.70833 15.0417C12.2061 15.0417 15.0417 12.2061 15.0417 8.70833C15.0417 5.21053 12.2061 2.375 8.70833 2.375C5.21053 2.375 2.375 5.21053 2.375 8.70833C2.375 12.2061 5.21053 15.0417 8.70833 15.0417Z" stroke="current" stroke-width="1.5" stroke-linecap="round" stroke-linejoin="round"/>
 <path d="M16.625 16.625L13.1812 13.1813" stroke="current" stroke-width="1.5" stroke-linecap="round" stroke-linejoin="round"/>
+
```
evulins: We could delete this line

src/components/molecules/SearchAutocomplete/index.test.tsx
```diff
@@ -126,32 +78,45 @@ describe("Input component", () => {
...
-          "Reptile",
-          "Solitary",
-          "Tail",
-          "Wetlands",
-        ]}
-      />
-    );
-    let input = screen.getByLabelText("Szukaj problemów") as HTMLInputElement;
+  it("should key press correctly", async () => {
+    testRenderer(<SearchAutocomplete setTag={() => {}} />);
```
evulins: Is there a better way to describe in test type of `setTag`?

src/components/molecules/SearchAutocomplete/index.tsx
```diff
@@ -1,22 +1,54 @@
...
+import {
+  GetSuggestionsQuery,
+  Suggestion,
+} from "../../organisms/ProblemSearchSection/interface";
 import SuggestionsList from "./suggestionsList";
 import * as S from "./styles";
 
-const SearchAutocomplete: FC<SearchAutocompleteProps> = ({ suggestions }) => {
-  const [filteredSuggestions, setFilteredSuggestions] = useState<string[]>([]);
+export const SUGGESTIONS = gql`
```
evulins: Do we want to keep this query in this file? Maybe we could move it to interface file?

src/components/molecules/SearchAutocomplete/index.tsx
```diff
@@ -29,22 +61,43 @@ const SearchAutocomplete: FC<SearchAutocompleteProps> = ({ suggestions }) => {
...
       if (activeSuggestionIndex !== -1) {
         e.preventDefault();
-        const test = filteredSuggestions[activeSuggestionIndex];
-        setSearchTerm(test);
-        setFilteredSuggestions([]);
+        const activeSuggestion =
+          filteredSuggestionsByType()[activeSuggestionIndex];
+        setSearchTerm(activeSuggestion.title);
+        setTag(activeSuggestion.title);
+        if (!loading && data) {
```
evulins: I'm wondering if we could move this code to some const and use it instead of writing the same code in so many places? What do you think?:

`if (!loading && data) {
      setFilteredSuggestions(data.suggestions);
    }`

src/components/pages/ProblemIndex/index.tsx
```diff
@@ -1,38 +1,57 @@
...
+
 import { primaryBackground } from "../../../assets/styles/colors";
 
 const ProblemIndex: FC = () => {
+  const { data } = useQuery<GetProblemsQuery>(PROBLEMS);
+
+  const allProblems = (data?.problems && data?.problems) || [];
+
+  const [tag, setTag] = useState<string>("");
+
```
evulins: I'm wondering if there is any rule to keep` useQuery `and` useState` closer together in the code or it doesn't really matter?

src/components/molecules/SearchAutocomplete/index.tsx
```diff
@@ -29,22 +61,43 @@ const SearchAutocomplete: FC<SearchAutocompleteProps> = ({ suggestions }) => {
...
+    if (!loading && data) {
+      setFilteredSuggestions(data.suggestions);
+    }
+    setSearchTerm(suggestion.title);
+    setTag(suggestion.title);
     setActiveSuggestionIndex(-1);
     setShowSuggestions(false);
   };
 
+  const handleKeyUp = (e: KeyboardEvent) => {
```
evulins: I don't think that `e: KeyboardEvent` is necessary here.

src/components/molecules/SearchAutocomplete/suggestionsList.tsx
```diff
@@ -17,14 +19,23 @@ const SuggestionsList: FC<SuggestionsProps> = ({
       aria-labelledby="search-input-label"
```
evulins: According to this article https://tink.uk/the-difference-between-aria-label-and-aria-labelledby/, we could use here just `aria-label`

src/components/molecules/SearchAutocomplete/suggestionsList.tsx
```diff
@@ -17,14 +19,23 @@ const SuggestionsList: FC<SuggestionsProps> = ({
...
       role="listbox"
     >
-      {filteredSuggestions.map((suggestion: string, index: number) => (
+      {filteredSuggestions.map((suggestion, index) => (
         <S.Suggestion
-          key={suggestion}
+          key={suggestion.id}
```
evulins, line 23: We should style differently suggestion for mobile devices - https://www.figma.com/file/FZKtRjUFs3QHXZgQKnJRMW/Thinking-Classroom?node-id=891%3A0  

We could do this also in different task.

src/components/pages/ProblemIndex/index.tsx
```diff
@@ -1,38 +1,57 @@
...
 import SearchResultTile from "../../molecules/SearchResultTile";
+import ProblemSearchSection from "../../organisms/ProblemSearchSection";
+import { GetProblemsQuery, PROBLEMS } from "./interface";
+
 import { primaryBackground } from "../../../assets/styles/colors";
 
 const ProblemIndex: FC = () => {
+  const { data } = useQuery<GetProblemsQuery>(PROBLEMS);
+
+  const allProblems = (data?.problems && data?.problems) || [];
```
bazelajoanna: Unnecessary `&& data?.problems`

evulins (commented):
Awesome work 🍾

src/components/molecules/SearchAutocomplete/index.tsx
```diff
@@ -1,22 +1,54 @@
...
 
+  useEffect(() => {
+    if (!loading && data) {
+      setFilteredSuggestions(data.suggestions);
+    }
+  }, [data]);
+
+  const filteredSuggestionsByType = () => {
+    const filteredSuggestionsCopy = [...filteredSuggestions];
+    filteredSuggestionsCopy.sort((a, b) =>
```
bazelajoanna: We can also improve sorting so that all tags will show in an alphabetically way, starting with given letter (in this example when we type "c", the first suggestion doesn't start from "c")

image

bazelajoanna (commented):
🥇 that looks nice! 😸

src/apolloClient.ts
franiakat: No :)

src/components/atoms/Tag/interface.ts
```diff
@@ -1,4 +1,9 @@
 export interface TagProps {
   text: string;
   isDroppable?: boolean;
+  fontSize?: string;
```
franiakat, line 4: Do we want to have `TagStyleProps` maybe?

src/components/molecules/SearchAutocomplete/index.tsx
franiakat: I think it's worth to move it elsewhere, but not sure whether interface is the place. 🤔

src/components/molecules/SearchAutocomplete/index.tsx
```diff
@@ -29,22 +61,43 @@ const SearchAutocomplete: FC<SearchAutocompleteProps> = ({ suggestions }) => {
...
 
-  const chooseSuggestion = (suggestion: string) => {
-    setFilteredSuggestions([]);
-    setSearchTerm(suggestion);
+  const chooseSuggestion = (suggestion: Suggestion) => {
+    if (!loading && data) {
+      setFilteredSuggestions(data.suggestions);
+    }
+    setSearchTerm(suggestion.title);
+    setTag(suggestion.title);
```
franiakat, line 59: Shouldn't we check the suggestion type before setting the tag?

src/components/molecules/SearchAutocomplete/suggestionsList.tsx
franiakat: let's do it separately :)

src/components/pages/ProblemIndex/index.tsx
franiakat: I'd keep it close, more of a style thing than rule though. ;)

src/components/molecules/SearchAutocomplete/index.tsx
bazelajoanna: I've created a new function `updateSuggestion` and put that code into it :)

src/components/molecules/SearchAutocomplete/suggestionsList.tsx
bazelajoanna: Yes, it is for tablets. I've created a new task in Asana :)

src/components/molecules/SearchAutocomplete/index.tsx
bazelajoanna: This function is used in ProblemIndex to return all problems that have this string among tags. So we don't need to know what suggestion type it is. We even don't use SUGGESTION Query here, but PROBLEMS.

src/components/molecules/SearchAutocomplete/index.tsx
bazelajoanna: Should we create a new task for that?

src/components/molecules/SearchAutocomplete/index.test.tsx
bazelajoanna: good question ;)

src/components/molecules/SearchAutocomplete/suggestionsList.tsx
bazelajoanna: Right, I've changed that :)

src/components/molecules/SearchAutocomplete/index.tsx
evulins: Previously we moved this schema to interface. For now, I would move this one also to interface. If we find a better place in the future, we can refactor.

src/components/molecules/SearchAutocomplete/index.tsx
evulins: Done :)

src/components/atoms/Tag/index.tsx
```diff
@@ -1,11 +1,11 @@
 import { FC } from "react";
 
 import * as S from "./styles";
-import { TagProps } from "./interface";
+import { TagStyleProps } from "./interface";
 
-const Tag: FC<TagProps> = ({ text, isDroppable = false }) => {
+const Tag: FC<TagStyleProps> = ({ text, isDroppable = false, fontSize }) => {
```
franiakat: Props with default should be always at the end, this way we can easily omit them, when we want to use defaults, just writing: `<Tag text="abs" fontSize="20px" />`
```suggestion
const Tag: FC<TagStyleProps> = ({ text, fontSize, isDroppable = false, }) => {
```

franiakat: Gonna fix it already 👯

franiakat: Actually does not matter, as this is not a function, but a functional component 😂

src/components/molecules/SearchAutocomplete/index.test.tsx
franiakat: Type or it's definition? :) I have just committed using `jest.fn()` here instead.

PR #22, opened 2021-12-13
Fix nav menu making page horizontally scrollable

bazelajoanna:
Change NavbarMenuList width to prevent horizontally scrolling.


franiakat (approved):
Nice!!!
sqash

PR #25, opened 2021-12-14
Fix fonts changes

evulins:
Fix font-changes on re-rendering components and change font types:

- Article about font types and using font-face: https://css-tricks.com/snippets/css/using-font-face/
- Post on styled-component repository about this issue https://github.com/styled-components/styled-components/issues/2227


franiakat (approved):
sqash

PR #26, opened 2021-12-14
Zero results view

bazelajoanna:
Add view that is displayed when there is no search result.

### DIFF
--- a/src/assets/styles/colors.js
+++ b/src/assets/styles/colors.js
@@ -10,15 +10,15 @@ const darkGradientBackground =
 const lightGradientBackground =
   "linear-gradient(93.51deg, #1782DC 8.9%, #6AB7F7 94.31%)";
 
-const activeButton = "#3678E3";
+const blue100 = "#3678E3";
 
-const hoverButton = "#124CAB";
+const blue200 = "#124CAB";
+
+const blue300 = "#161C56";
 
 const disabledButton = "#C9DAF6";
 
 const hoverSecondaryButton = "#E5EFFF";
-
-const activeLink = "#3678E3";
 
 const placeholder = "#A8A8A8";
 
@@ -29,10 +29,6 @@ const successMessage = "#058C23";
 const tagText = "#DA7602";
 
 const tagBackground = "#FFF1DE";
-
-const mainHeader = "#161C56";
-
-const primarySubtitle = "#124CAB";
 
 const secondarySubtitle = "#F88B0C";
 
@@ -46,18 +42,16 @@ export {
   primaryBackground,
   darkGradientBackground,
   lightGradientBackground,
-  activeButton,
-  hoverButton,
+  blue100,
+  blue200,
+  blue300,
   disabledButton,
   hoverSecondaryButton,
-  activeLink,
   placeholder,
   errorMessage,
   successMessage,
   tagText,
   tagBackground,
-  mainHeader,
-  primarySubtitle,
   secondarySubtitle,
   lightGrey,
   textDetails,
--- a/src/components/atoms/ResultCountLabel/index.tsx
+++ b/src/components/atoms/ResultCountLabel/index.tsx
@@ -2,11 +2,14 @@ import { FC } from "react";
 
 import * as S from "./styles";
 import { ResultCountWrapperProps } from "./interface";
+import { white, primaryBackground } from "../../../assets/styles/colors";
 
 const ResultCountLabel: FC<ResultCountWrapperProps> = ({ count }) => {
   return (
     <S.ResultCountWrapper>
-      <S.ResultCountP>Wyniki wyszukiwania ({count})</S.ResultCountP>
+      <S.ResultCountP backgroundColor={count === 0 ? white : primaryBackground}>
+        Wyniki wyszukiwania ({count})
+      </S.ResultCountP>
     </S.ResultCountWrapper>
   );
 };
--- a/src/components/atoms/ResultCountLabel/index.tsx
+++ b/src/components/atoms/ResultCountLabel/index.tsx
@@ -7,9 +7,11 @@ import { white, primaryBackground } from "../../../assets/styles/colors";
 const ResultCountLabel: FC<ResultCountWrapperProps> = ({ count }) => {
   return (
     <S.ResultCountWrapper>
-      <S.ResultCountP backgroundColor={count === 0 ? white : primaryBackground}>
+      <S.ResultCountSpan
+        backgroundColor={count === 0 ? white : primaryBackground}
+      >
         Wyniki wyszukiwania ({count})
-      </S.ResultCountP>
+      </S.ResultCountSpan>
     </S.ResultCountWrapper>
   );
 };
--- a/src/components/atoms/ResultCountLabel/styles.tsx
+++ b/src/components/atoms/ResultCountLabel/styles.tsx
@@ -1,8 +1,7 @@
 import styled from "styled-components";
 
 import Wrapper from "../Wrapper";
-import P from "../P";
-import { ResultCountPProps } from "./interface";
+import { ResultCountSpanProps } from "./interface";
 import { minDevice } from "../../../assets/styles/breakpoints";
 import { textDetails, primaryBackground } from "../../../assets/styles/colors";
 
@@ -36,7 +35,7 @@ const ResultCountWrapper = styled(Wrapper)`
   }
 `;
 
-const ResultCountP = styled(P)<ResultCountPProps>`
+const ResultCountSpan = styled.span<ResultCountSpanProps>`
   line-height: 2.4rem;
   background-color: ${(props) => props.backgroundColor};
   height: 2.4rem;
@@ -49,4 +48,4 @@ const ResultCountP = styled(P)<ResultCountPProps>`
     background-color: ${primaryBackground};
   }
 `;
-export { ResultCountWrapper, ResultCountP };
+export { ResultCountWrapper, ResultCountSpan };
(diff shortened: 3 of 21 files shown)


src/assets/styles/colors.js
```diff
@@ -10,16 +10,16 @@ const darkGradientBackground =
 const lightGradientBackground =
   "linear-gradient(93.51deg, #1782DC 8.9%, #6AB7F7 94.31%)";
 
-const activeButton = "#3678E3";
+const blue100 = "#3678E3";
 
-const hoverButton = "#124CAB";
+const blue200 = "#124CAB";
+
+const blue300 = "#161C56";
```
franiakat, line 17: 🙆‍♀️

src/components/molecules/SearchAutocomplete/index.tsx
```diff
@@ -5,18 +5,23 @@ import { SearchAutocompleteProps, SUGGESTIONS } from "./interface";
...
 import * as S from "./styles";
 import { sortSuggestions } from "../../pages/ProblemIndex/utils";
 
-const SearchAutocomplete: FC<SearchAutocompleteProps> = ({ setTag }) => {
+const SearchAutocomplete: FC<SearchAutocompleteProps> = ({
+  setTag,
+  setProblems,
+  setSearchTerm,
+  searchTerm,
+}) => {
```
evulins, line 20: Those propos  should also be added to test component.

 ```
 setProblems,
  setSearchTerm,
  searchTerm,
```

File: src/components/molecules/ZeroResults/index.tsx [hunk 3]
```diff
@@ -0,0 +1,32 @@
...
+import { ZeroResultsProps } from "./interface";
+
+import * as S from "./styles";
+
+const ZeroResults: FC<ZeroResultsProps> = ({ setTag, setSearchTerm }) => {
+  const suggestions = ["Ciągi", "Jedzenie", "Zwierzęta"];
+
+  const handleOnClick = (el: string) => {
+    setTag(el);
+    setSearchTerm(el);
```
evulins: We  could use one style of writing **element** We can choose which one. I don't have favourite one :)

evulins (commented):
Looks great! 👯

src/components/atoms/ResultCountLabel/index.tsx
```diff
@@ -2,11 +2,14 @@ import { FC } from "react";
 
 import * as S from "./styles";
 import { ResultCountWrapperProps } from "./interface";
+import { white, primaryBackground } from "../../../assets/styles/colors";
 
 const ResultCountLabel: FC<ResultCountWrapperProps> = ({ count }) => {
   return (
     <S.ResultCountWrapper>
-      <S.ResultCountP>Wyniki wyszukiwania ({count})</S.ResultCountP>
+      <S.ResultCountP backgroundColor={count === 0 ? white : primaryBackground}>
```
franiakat: I cannot see any difference in the background colour when count is 0 or non 0. 🤔

src/components/atoms/ResultCountLabel/styles.tsx
```diff
@@ -35,13 +36,17 @@ const ResultCountWrapper = styled(Wrapper)`
   }
 `;
 
-const ResultCountP = styled(P)`
+const ResultCountP = styled(P)<ResultCountPProps>`
```
franiakat: It is not a paragraph, should be rather span I think ;)

src/components/molecules/ZeroResults/index.test.tsx
```diff
@@ -0,0 +1,27 @@
...
+  const setSearchTerm = jest.fn();
+  it("displays image", async () => {
+    testRenderer(<ZeroResults setTag={setTag} setSearchTerm={setSearchTerm} />);
+
+    expect(
+      await screen.findByLabelText("Dziewczyna patrząca przez lornetkę")
+    ).toBeInTheDocument();
+
+    let jedzenie = await screen.findByText(/Jedzenie/i);
+    fireEvent.click(jedzenie);
```
franiakat:
```suggestion
    let tag = await screen.findByText(/Jedzenie/i);
    fireEvent.click(tag);
```

src/components/molecules/ZeroResults/index.tsx
franiakat: The other one would be `element`? Actually here I'd say that we have `tag` or some sort of `tagTitle` simply ;)

src/components/molecules/ZeroResults/index.tsx
```diff
@@ -0,0 +1,32 @@
...
+  };
+
+  return (
+    <S.ZeroResultsWrapper>
+      <S.ZeroResultsImg aria-label="Dziewczyna patrząca przez lornetkę" />
```
franiakat: I'd skip `aria-label` and would actually hide this decoration element from screen reader ;)
```suggestion
      <S.ZeroResultsImg aria-hidden="true" />
```

src/components/molecules/ZeroResults/index.tsx
```diff
@@ -0,0 +1,32 @@
...
+      <S.ZeroResultsHeader>Brak wyników</S.ZeroResultsHeader>
+      <S.ZeroResultsP>Spróbuj wyszukać:</S.ZeroResultsP>
```
franiakat: It's not a paragraph, a span or tiny heading ;)

src/components/molecules/ZeroResults/index.tsx
```diff
@@ -0,0 +1,32 @@
+import { FC } from "react";
+import { ZeroResultsProps } from "./interface";
+
+import * as S from "./styles";
+
+const ZeroResults: FC<ZeroResultsProps> = ({ setTag, setSearchTerm }) => {
+  const suggestions = ["Ciągi", "Jedzenie", "Zwierzęta"];
```
franiakat:
```suggestion
  const suggestedTags = ["Ciągi", "Jedzenie", "Zwierzęta"];
```

src/components/molecules/ZeroResults/index.tsx
```diff
@@ -0,0 +1,32 @@
...
+    setSearchTerm(el);
+  };
+
+  return (
+    <S.ZeroResultsWrapper>
+      <S.ZeroResultsImg aria-label="Dziewczyna patrząca przez lornetkę" />
+      <S.ZeroResultsHeader>Brak wyników</S.ZeroResultsHeader>
+      <S.ZeroResultsP>Spróbuj wyszukać:</S.ZeroResultsP>
+      <S.ZeroResultsSuggestions>
+        {suggestions.map((suggestion, id) => (
```
franiakat:
```suggestion
        {suggestedTag.map((tag, id) => (
```

src/components/molecules/ZeroResults/index.tsx
```diff
@@ -0,0 +1,32 @@
...
+    setTag(el);
+    setSearchTerm(el);
+  };
+
+  return (
+    <S.ZeroResultsWrapper>
+      <S.ZeroResultsImg aria-label="Dziewczyna patrząca przez lornetkę" />
+      <S.ZeroResultsHeader>Brak wyników</S.ZeroResultsHeader>
+      <S.ZeroResultsP>Spróbuj wyszukać:</S.ZeroResultsP>
+      <S.ZeroResultsSuggestions>
```
franiakat:
```suggestion
      <S.ZeroResultsTagSuggestions>
```

src/components/pages/ProblemIndex/index.test.tsx
```diff
@@ -35,4 +35,20 @@ describe("<ProblemIndex/>", () => {
       await screen.findByText("Wyniki wyszukiwania (2)")
     ).toBeInTheDocument();
   });
+
+  xit("displays zero results view when result count = 0", async () => {
```
franiakat, line 39: 😓

franiakat (changes requested):
Please see a few suggestions :))

src/components/atoms/ResultCountLabel/index.tsx
bazelajoanna: According to design, background colour changes only on mobile. For tablets and wider screens there is no difference.

src/components/molecules/ZeroResults/index.tsx
bazelajoanna: Ok, I've changed `el` to `tag`.

src/components/atoms/ResultCountLabel/styles.tsx
bazelajoanna: Good point, I've changed it :)

src/components/molecules/ZeroResults/index.tsx
bazelajoanna: Right, I've changed it :)

src/components/atoms/ResultCountLabel/index.tsx
franiakat: Yes, but why should it depend on the count? If I understand it correctly it should be connected with the viewport size (mobile).

franiakat: Nvm, I've misread the design. ;)

src/components/molecules/ZeroResults/index.tsx
franiakat: @bazelajoanna

bazelajoanna: Done ;)

franiakat (approved):
sqash

PR #27, opened 2021-12-15
Fix styles for large devices

bazelajoanna:
Change font size for large devices in suggestion list and result count label.


src/components/atoms/ResultCountLabel/styles.tsx
```diff
@@ -37,11 +37,21 @@ const ResultCountWrapper = styled(Wrapper)`
 
 const ResultCountP = styled(P)`
   line-height: 2.4rem;
+  font-size: 1.4rem;
```
franiakat, line 40: 😻

franiakat (approved):
sqash

PR #28, opened 2021-12-15
Fix style input value

evulins:
Allows displaying tag in the input field.


src/components/atoms/Tag/index.tsx
```diff
@@ -1,19 +1,25 @@
...
+  text,
+  fontSize,
+  onRemoveTag,
+  isDroppable = false,
+}) => (
+  <S.TagWrapper fontSize={fontSize}>
+    {text}
+    {isDroppable && (
+      <S.RemoveButton
+        aria-label="Usuń tag"
```
franiakat, line 16: 👏

src/components/atoms/Tag/interface.ts
```diff
@@ -1,9 +1,19 @@
...
+};
+
+type DroppableTag = {
+  text: string;
+  isDroppable?: true;
+  fontSize?: string;
+  onRemoveTag?: () => void;
+};
+
+export type TagProps = Tag | DroppableTag;
```
franiakat:
```
type CommonTagProps = {
  text: string;
  fontSize?: string;
};

type Tag = {
  isDroppable?: false;
  onRemoveTag?: never;
};

type DroppableTag = {
  isDroppable?: true;
  onRemoveTag?: () => void;
};

export type TagProps = (Tag | DroppableTag) & CommonTagProps;
```

src/components/molecules/SearchAutocomplete/index.test.tsx
```diff
@@ -1,5 +1,6 @@
 import { screen, fireEvent, within } from "@testing-library/react";
 import userEvent from "@testing-library/user-event";
+import { debug } from "util";
```
franiakat: leftover ;)

src/components/molecules/SearchAutocomplete/index.test.tsx
```diff
@@ -45,16 +46,16 @@ describe("Input component", () => {
     expect(input.value).toBe("");
     expect(screen.queryByRole("listbox")).not.toBeInTheDocument();
 
-    fireEvent.change(input, { target: { value: "Oce" } });
-    expect(input).toHaveValue("Oce");
+    fireEvent.change(input, { target: { value: "Ciągi" } });
```
franiakat: You can also use [userEvent lib](https://testing-library.com/docs/ecosystem-user-event/#typeelement-text-options), it's not a crazy event, but the syntax is more concise ;)

src/components/molecules/SearchAutocomplete/index.test.tsx
```diff
@@ -120,4 +121,51 @@ describe("Input component", () => {
...
+  it("should display input with tag as a value and return to default input after removing tag", async () => {
+    testRenderer(<SearchAutocomplete setTag={setTag} />);
+    let input = (await screen.findByLabelText(
+      "Szukaj problemów"
+    )) as HTMLInputElement;
+
+    fireEvent.change(input, { target: { value: "J" } });
+    expect(screen.queryByText("Jedzenie")).toBeInTheDocument();
+    fireEvent.keyDown(input, { key: "ArrowDown", code: 40 });
+    fireEvent.keyDown(input, { key: "Enter", code: 13 });
```
franiakat: I think the code can be skipped here.

src/components/molecules/SearchAutocomplete/index.test.tsx
```diff
@@ -120,4 +121,51 @@ describe("Input component", () => {
...
+
+    const removeTagButton = (await screen.findByRole("button", {
+      name: /Usuń tag/i,
+    })) as HTMLButtonElement;
+    expect(removeTagButton).toBeInTheDocument();
+    fireEvent.click(removeTagButton);
+
+    let defaultInput = (await screen.findByLabelText(
```
franiakat: You can use `const` here.

src/components/molecules/SearchAutocomplete/index.tsx
```diff
@@ -18,6 +19,7 @@ const SearchAutocomplete: FC<SearchAutocompleteProps> = ({ setTag }) => {
   const [showSuggestions, setShowSuggestions] = useState(false);
   const [searchTerm, setSearchTerm] = useState("");
   const [showClearButton, setShowClearButton] = useState(false);
+  const [suggestionType, setSuggestionType] = useState("");
```
franiakat: I think we don't need additional state here (`[suggestionType, setSuggestionType]`), I'd set tag only when suggestion id of type TAG and therefore whenever we check whether tag exists we can assume the suggestion is of type TAG -> we use our fake input with proper styling. I have checked it locally and works good :)

src/components/molecules/SearchAutocomplete/index.tsx
```diff
@@ -106,34 +112,47 @@ const SearchAutocomplete: FC<SearchAutocompleteProps> = ({ setTag }) => {
...
-          value={searchTerm}
-          onChange={handleChange}
-          onKeyDown={handleKeyDown}
-          onKeyUp={handleKeyUp}
-          onSubmit={(e) => {
-            e.preventDefault();
-          }}
-          showSuggestions={showSuggestions}
-        />
+        {searchTerm && suggestionType === "tag" ? (
```
franiakat: Here we'd have 
```suggestion
        {searchTerm && !!tag ? (
```
and `tag` would come from a parent altogether with `setTag`

src/components/molecules/SearchAutocomplete/index.tsx
```diff
@@ -106,34 +112,47 @@ const SearchAutocomplete: FC<SearchAutocompleteProps> = ({ setTag }) => {
...
+          <>
+            <S.Button aria-label="Remove button" onClick={() => clearInput()}>
+              <span id="Remove button" hidden>
+                Remove input text button
+              </span>
 
-            <S.RemoveIcon aria-hidden="true" />
-          </S.Button>
+              <S.RemoveIcon aria-hidden="true" />
+            </S.Button>
```
franiakat: hm, I think I wouldn't keep it for tag inputs ;)

src/components/molecules/SearchAutocomplete/inputWithTags.tsx
```diff
@@ -0,0 +1,36 @@
...
+  onRemoveTag,
+  setSuggestionType,
+}) => {
+  const handleKeyDown = (e: React.KeyboardEvent<HTMLSpanElement>) => {
+    if (e.key === "Backspace") {
+      setSuggestionType("");
+    }
+    if (e.key === "Enter") {
+      e.preventDefault();
+      return false;
```
franiakat: we could use switch case here ;)

src/components/molecules/SearchAutocomplete/interface.ts
```diff
@@ -19,6 +19,16 @@ export interface InputStyleProps {
   showSuggestions: boolean;
 }
 
+export interface InputWithTagsProps {
+  onRemoveTag: () => void;
+  text: string;
+  setSuggestionType: (type: string) => void;
```
franiakat: I'd keep the functions at the end, but it's just a nitpick and a matter of style. ;)

src/components/molecules/SearchAutocomplete/styles.tsx
```diff
@@ -63,15 +57,29 @@ const Input = styled.input<InputStyleProps>`
...
+  border-radius: ${(props) => (props.showSuggestions ? 0 : "5px")};
+  border-top-left-radius: 5px;
+  border-top-right-radius: 5px;
+  ::placeholder {
+    color: ${placeholder};
+  }
+  &:focus-visible {
+    outline: none;
+  }
+  ${inputStyles}
```
franiakat, line 70: nice!

franiakat (changes requested):

src/components/molecules/SearchAutocomplete/index.test.tsx
evulins: How can I trigger to show input with tags then?

src/components/molecules/SearchAutocomplete/index.test.tsx
```diff
@@ -120,4 +122,51 @@ describe("Input component", () => {
...
+
+    fireEvent.change(input, { target: { value: "J" } });
+    expect(screen.queryByText("Jedzenie")).toBeInTheDocument();
+    fireEvent.keyDown(input, { key: "ArrowDown", code: 40 });
+    fireEvent.keyDown(input, { key: "Enter", code: 13 });
+
+    const removeTagButton = await screen.findByRole("button", {
+      name: /Usuń tag/i,
+    });
+    expect(removeTagButton).toBeInTheDocument();
```
bazelajoanna: Do we need to check this in this test case?

franiakat: No, we can remove it on the branch with tests fixing.

src/components/molecules/SearchAutocomplete/index.test.tsx
franiakat: We can also remove this on a branch with tests. :)

franiakat (approved):

franiakat:
sqash

PR #29, opened 2021-12-21
Fix Integration tests for Problem Index functionality

franiakat:
Fix integration tests for Problem Index page.

### DIFF
--- a/src/components/molecules/SearchAutocomplete/inputWithTags.tsx
+++ b/src/components/molecules/SearchAutocomplete/inputWithTags.tsx
@@ -5,28 +5,21 @@ import Tag from "../../atoms/Tag";
 import * as S from "./styles";
 import { InputWithTagsProps } from "./interface";
 
-const InputWithTags: FC<InputWithTagsProps> = ({
-  text,
-  onRemoveTag,
-  setTag,
-}) => {
+const InputWithTags: FC<InputWithTagsProps> = ({ text, onRemoveTag }) => {
   const handleKeyDown = (e: React.KeyboardEvent<HTMLSpanElement>) => {
-    switch (e.key) {
-      case "Backspace":
-        break;
-      case "Enter":
-        e.preventDefault();
-        break;
-      default:
+    if (e.key === "Backspace") {
+      onRemoveTag();
+    } else if (e.key === "Enter") {
+      e.preventDefault();
     }
   };
 
   return (
     <S.InputWithTags
-      aria-label="Input z tagiem"
+      aria-label="Input filtrujący po tagu"
       contentEditable="true"
       suppressContentEditableWarning={true}
-      onKeyDown={(e) => handleKeyDown(e)}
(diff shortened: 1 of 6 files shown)


src/components/pages/ProblemIndex/index.test.tsx
```diff
@@ -1,17 +1,138 @@
...
+  it("should show suggestions list matching input value", async () => {
+    testRenderer(<ProblemIndex />);
+
+    const input = (await screen.findByLabelText(
+      "Szukaj problemów"
+    )) as HTMLInputElement;
+    expect(input.value).toBe("");
+    expect(screen.queryByRole("listbox")).not.toBeInTheDocument();
+
+    fireEvent.change(input, { target: { value: "c" } });
```
evulins: Do we want in every place like this to use userEvent instead of fireEvent? Or just in some cases?

evulins (approved):
👏 👏 👏 
I only have philosophical questions - one I ask about userEvent and the second one where now we should component inputWithTags?

src/components/pages/ProblemIndex/index.test.tsx
franiakat: Everywhere, that's a good catch! :) I will fix it everywhere.
It's better to use userEvent package as it is closer to actual user experience, here it's a nice and short [explanation by Kent C. Dodds](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#not-using-testing-libraryuser-event)

franiakat:
@evulins 
> where now we should component inputWithTags?
What do you mean? Where should we test it? :)

evulins:
>  What I meant is that I wrote two tests in the previous test file for the pseudo inputWithTags component.  And because of our mechanism structure they didn't work properly. Right now they are missing :D And I wonder if it's on purpose and if yes should we test it somewhere else?

franiakat:
@evulins oh, sure, I just forgot, I will paste them here, as because of the mechanics they also belong here. :)

bazelajoanna:
👌 Nice!

src/components/molecules/SearchAutocomplete/inputWithTags.tsx
```diff
@@ -5,28 +5,21 @@ import Tag from "../../atoms/Tag";
...
-      case "Backspace":
-        break;
-      case "Enter":
-        e.preventDefault();
-        break;
-      default:
+    if (e.key === "Backspace") {
+      onRemoveTag();
+    } else if (e.key === "Enter") {
+      e.preventDefault();
```
franiakat, line 13: Oh I forgot to change it back to switch case, it was easier to debug this way.

franiakat: I will leave it this way, as I started to break one test case with this `switch case` "refactor" 🙈

bazelajoanna (approved):
👏

PR #30, opened 2021-12-22
Add routing

bazelajoanna:
Add React Router v6 using documentation: 
https://reactrouter.com/docs/en/v6/getting-started/installation


src/App.tsx
```diff
@@ -1,14 +1,25 @@
...
     <>
       <Navbar />
-      <ProblemIndex />
+      <Routes>
+        {/* <Route path="/" element {<LandingPage/>}> */}
+        <Route path="/" element={<ProblemIndex />} />
+        <Route path="/problems/:problemId" element={<DetailedProblemView />} />
+        <Route path="/about" element={<AboutMethod />} />
+        <Route path="/moderator" element={<Moderator />} />
+      </Routes>
```
franiakat: 👏

src/assets/styles/global-styles.js
```diff
@@ -59,8 +59,14 @@ ${fonts}
   a {
     text-decoration: none;
     transition: color 0.05s ease-in-out;
+
```
franiakat: what about not visited link colour? :)

src/components/molecules/AboutMethod/index.tsx
```diff
@@ -0,0 +1,14 @@
+import { FC } from "react";
+import { Link } from "react-router-dom";
+
+const AboutMethod: FC = () => {
```
franiakat: How about we have a convention that we keep those views/pages in `pages/` directory? We could also agree on a naming convention, e.g.:
`AboutMethodView` or `AboutMethodPage` and align all page-like components also `ProblemIndexPage` or accordingly `ProblemIndexView`.
Btw I would rename it to `ProblemsView` or `ProblemsPage`.

src/components/molecules/SearchResultTile/index.tsx
```diff
@@ -27,9 +31,11 @@ const SearchResultTile: FC<ProblemSummaryProps> = ({
...
         </S.TagWrapper>
       </S.DetailsWrapper>
-      <Button isPrimary={false} isHidden>
-        Zobacz treść
-      </Button>
+      <Link to={`/problems/${id}`}>
+        <Button isPrimary={false} isHidden>
+          Zobacz treść
+        </Button>
+      </Link>
```
franiakat: I think ideal would be that we have Link from react router that is styled as component, as having button tag inside of a link tag seems a bit odd. ;)

src/components/organisms/Navbar/index.tsx
```diff
@@ -48,8 +51,12 @@ const Navbar: FC = () => {
...
       <S.NavbarMenu className={navbarMenuStyle}>
         <S.NavbarMenuList>
-          <S.NavbarMenuItem>O metodzie</S.NavbarMenuItem>
-          <S.NavbarMenuItem>Strefa moderatora</S.NavbarMenuItem>
+          <Link to="/about">
+            <S.NavbarMenuItem>O metodzie</S.NavbarMenuItem>
+          </Link>
+          <Link to="/moderator">
+            <S.NavbarMenuItem>Strefa moderatora</S.NavbarMenuItem>
+          </Link>
```
franiakat: It should be the other way around `li` tags should contain `a` tags.

src/components/pages/DetailedProblemView/index.tsx
```diff
@@ -0,0 +1,44 @@
+import { FC, useEffect, useState } from "react";
+import { Link, useParams } from "react-router-dom";
+import { useQuery } from "@apollo/client";
+
+import { GetProblemsQuery, PROBLEMS } from "../ProblemIndex/interface";
+import { ProblemSummaryProps } from "../../molecules/SearchResultTile/interface";
+
+const DetailedProblemView: FC = () => {
```
franiakat: ProblemDetailedView would be more correct in English, I think 🙃

src/components/pages/DetailedProblemView/index.tsx
```diff
@@ -0,0 +1,44 @@
...
+  const params = useParams();
+  const id: any = params.problemId;
+
+  const { data, loading } = useQuery<GetProblemsQuery>(PROBLEMS);
```
franiakat: We can rethink it, when implementing this page precisely, but we need for sure additional call to the API to give us more details about the page, but we should treat it as a separate task, so let's not change the approach here and now. ;)

src/components/pages/DetailedProblemView/index.tsx
```diff
@@ -0,0 +1,44 @@
+import { FC, useEffect, useState } from "react";
+import { Link, useParams } from "react-router-dom";
+import { useQuery } from "@apollo/client";
+
+import { GetProblemsQuery, PROBLEMS } from "../ProblemIndex/interface";
+import { ProblemSummaryProps } from "../../molecules/SearchResultTile/interface";
+
+const DetailedProblemView: FC = () => {
+  const params = useParams();
+  const id: any = params.problemId;
```
franiakat: `number`?
```suggestion
  const id: number = params.problemId;
```

src/components/pages/DetailedProblemView/index.tsx
```diff
@@ -0,0 +1,44 @@
...
+      const currentProblem = problems[id - 1];
+      setProblems(filteredProblems);
+      setProblemDetails({
+        title: currentProblem.title,
+        author: currentProblem.author,
+      });
+    }
+  }, [data?.problems.length, loading]);
+
+  return data?.problems && !loading ? (
```
franiakat: It's used twice, can be extracted to a constant.

franiakat (changes requested):
Awesome job, Asia! Happy to see new possible interactions in our app. 🍾

src/assets/styles/global-styles.js
bazelajoanna: I've removed `color: inherit` from visited link and moved it to an `a` element. Now it works correctly :)

src/components/molecules/AboutMethod/index.tsx
bazelajoanna: Good point, I've renamed them to "...`Page`" and moved to `pages/` directory :)

src/components/pages/DetailedProblemView/index.tsx
bazelajoanna: At first I tried with `number`, but the error occurred. That's why i set `any`. Now when I set as `number` there is the same message: `Type 'string | undefined' is not assignable to type 'number'.
  Type 'undefined' is not assignable to type 'number'`

src/components/pages/DetailedProblemView/index.tsx
bazelajoanna: Do you mean `data?.problems`? Or the whole statement `data?.problems && !loading`?

src/components/pages/DetailedProblemView/index.tsx
franiakat: I see, yes params would be strings always, so here I'd go with
```suggestion
  const id: string | undefined = params.problemId;
```

src/components/pages/DetailedProblemView/index.tsx
franiakat: I meant `data?.problems && !loading`, so basically you'd have sth along those lines:
`const problemsLoaded: boolean = data?.problems && !loading;`

bazelajoanna: Done, thanks :)

src/components/pages/DetailedProblemView/index.tsx
bazelajoanna: I've changed it to: `const id: string = params.problemId || "";`

franiakat (approved):
Really good work, Asia, I am happy to approve 🙃 👏
sqash

PR #31, opened 2021-12-28
Add scroll to top button

evulins:
Create scroll to the button which will be added to the show problem page


src/components/pages/ProblemsPage/index.tsx
```diff
@@ -63,6 +64,7 @@ const ProblemsPage: FC = () => {
           ))
         )}
       </Wrapper>
+      <ScrollToTopButton />
```
franiakat, line 67: In the end this won't be added on this page, but on ProblemDetailedPage, right? :)

src/components/atoms/ScrollToTopButton/styles.tsx
```diff
@@ -0,0 +1,27 @@
...
+import { minDevice } from "../../../assets/styles/breakpoints";
+
+const ScrollButton = styled.button`
+  position: fixed;
+  left: 85%;
+  top: 10%;
+  width: fit-content;
+  background: none;
+  border: none;
+  overflow-y: scroll;
```
franiakat: Is that needed? I think not necessarily, but am I missing something? :)

src/components/atoms/ScrollToTopButton/styles.tsx
```diff
@@ -0,0 +1,27 @@
...
+`;
+
+const ArrowUpIcon = styled(ArrowUp)<ArrowUpIconStyleProps>`
+  display: ${(props) => props.display};
```
franiakat: Is this needed to be configurable from the outside? 🤔

src/components/atoms/ScrollToTopButton/index.tsx
```diff
@@ -0,0 +1,37 @@
+import { FC, useState } from "react";
+
+import * as S from "./styles";
+
+const ScrollToTopButton: FC = () => {
+  const [visible, setVisible] = useState(false);
+
+  const toggleVisible = () => {
```
franiakat, line 9: Right now this function is being called on every `scroll` event which is a lot, going up and down randomly on our problem index now gives me hundreds of them 🙃 Therefore we can throttle the execution of the function, in a simpler words, sort of delay it on given conditions: https://www.sitepoint.com/throttle-scroll-events/
You can take a look at this article here. ;)

src/components/atoms/ScrollToTopButton/index.test.tsx
```diff
@@ -0,0 +1,28 @@
+import { fireEvent, screen } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
+
+import { testRenderer } from "../../../setupTests";
+import ScrollToTopButton from "./index";
+
+describe("<ScrollToTopButtom />", () => {
+  it("doesn't display ScrollButton when the page is render", async () => {
```
franiakat: :)
```suggestion
  it("doesn't display ScrollButton when the page is first rendered", async () => {
```

src/components/atoms/ScrollToTopButton/index.test.tsx
```diff
@@ -0,0 +1,28 @@
...
+    testRenderer(<ScrollToTopButton />);
+
+    expect(
+      screen.queryByTitle(/Ikona przewijania strony do początku/i)
+    ).not.toBeVisible();
+  });
+
+  // I think we need to test this case in the component when the button will be implemented
```
franiakat, line 16: Do you mean used in ProblemDetailedPage? :)

src/components/atoms/ScrollToTopButton/index.test.tsx
```diff
@@ -0,0 +1,28 @@
+import { fireEvent, screen } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
+
+import { testRenderer } from "../../../setupTests";
+import ScrollToTopButton from "./index";
+
+describe("<ScrollToTopButtom />", () => {
```
franiakat, line 7: BTW around this topic of testing our new component, jest has its limitations (what we already know from other tests that can't be written here): https://github.com/testing-library/react-testing-library/issues/671
We may think of adding puppeteer or Cypress as KCD suggests 🙃

franiakat (changes requested):
Looking good! 💃 I had a few questions/thoughts. :)

franiakat:
Two additional thoughts:
- you can take a look at this article saying how to configure the css behaviour for people with `prefers-reduced-motion'` configuration
https://ashleemboyer.com/accessible-smooth-scroll-to-top-buttons-with-little-code
- the nasty DS_Store files got added 🙊 you can remove them.

src/components/pages/ProblemsPage/index.tsx
evulins: Of course, I put it here just to see and test the component :)

src/components/atoms/ScrollToTopButton/index.test.tsx
evulins: Yes :)

src/components/atoms/ScrollToTopButton/index.test.tsx
evulins: I saw this discussion earlier today (when i was googling how to test scrolling) and i want to talk about KCD suggestions

src/components/atoms/ScrollToTopButton/styles.tsx
evulins: This is this place when typescript was screaming at me :( We can discuss this during our daily :)

src/components/atoms/ScrollToTopButton/index.tsx
evulins: Good point. After I did some reading  and testing some functions I chose to add loadash.throttle to our app  :)

evulins:
> Two additional thoughts:
> 
> * you can take a look at this article saying how to configure the css behaviour for people with `prefers-reduced-motion'` configuration
>   https://ashleemboyer.com/accessible-smooth-scroll-to-top-buttons-with-little-code

We can manage focus when the button will be used in the ProblemDetailedPage

src/components/atoms/ScrollToTopButton/index.test.tsx
bazelajoanna: I think we can test it here, but we need to render `Navbar`, `ProblemsPage` and `Footer`:
`testRenderer(
      <>
        <Navbar />
        <ProblemsPage />
        <Footer />
      </>
    )`

src/components/atoms/ScrollToTopButton/index.tsx
```diff
@@ -0,0 +1,37 @@
...
+      <S.ArrowUpIcon
+        title="Ikona przewijania strony do początku"
+        onClick={scrollToTop}
+        display={visible ? "inline" : "none"}
+      />
+    </S.ScrollButton>
+  );
+};
+
+export default ScrollToTopButton;
```
bazelajoanna, line 37: 👏 👏 👏

src/components/atoms/ScrollToTopButton/index.test.tsx
evulins: Exactly ;) IMO this is the same case as it was with input. We don't want to render other components than the exact one we  would like test.

bazelajoanna (commented):
Very nice! 🎊

src/components/atoms/ScrollToTopButton/index.test.tsx
franiakat: Yes, let's wait until then and we will move it to ProblemDetailedPage test. 👌

franiakat (approved):
Very good piece of code 🎖️
sqash

PR #32, opened 2021-12-28
Add Subtitle component

evulins:
Add Problem Subtitle Header

### DIFF
--- a/src/components/atoms/ProblemSubtitle/index.test.tsx
+++ b/src/components/atoms/ProblemSubtitle/index.test.tsx
@@ -1,6 +1,6 @@
 import { screen } from "@testing-library/react";
-import { testRenderer } from "../../../setupTests";
 
+import { testRenderer } from "../../../setupTests";
 import ProblemSubtitle from "./index";
 
 describe("<ProblemSubtitle />", () => {
--- a/src/components/atoms/ProblemSubtitle/index.test.tsx
+++ b/src/components/atoms/ProblemSubtitle/index.test.tsx
@@ -1,0 +1,12 @@
+import { screen } from "@testing-library/react";
+import { testRenderer } from "../../../setupTests";
+
+import ProblemSubtitle from "./index";
+
+describe("<ProblemSubtitle />", () => {
+  it("displays problem subtitle", () => {
+    testRenderer(<ProblemSubtitle subtitle={"Podtytuł"} color={"blue"} />);
+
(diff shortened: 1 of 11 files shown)


src/components/atoms/ProblemSubtitle/index.test.tsx
```diff
@@ -0,0 +1,12 @@
+import { screen } from "@testing-library/react";
+import { testRenderer } from "../../../setupTests";
+
+import ProblemSubtitle from "./index";
```
franiakat:
```suggestion
import { screen } from "@testing-library/react";

import { testRenderer } from "../../../setupTests";
import ProblemSubtitle from "./index";
```

src/components/atoms/ProblemSubtitle/interface.ts
```diff
@@ -0,0 +1,8 @@
+export interface ProblemSubtitleProps {
+  subtitle: string;
+  color?: string;
+}
+
+export interface SubtitleHeaderStyleProps {
+  color?: string;
+}
```
franiakat:
```suggestion
export interface ProblemSubtitleProps extends SubtitleHeaderStyleProps {
  subtitle: string;
}

export interface SubtitleHeaderStyleProps {
  color?: string;
}
```

src/components/atoms/ProblemSubtitle/styles.tsx
```diff
@@ -0,0 +1,47 @@
+import styled from "styled-components";
+
+import { SubtitleHeaderStyleProps } from "./interface";
+import Wrapper from "../Wrapper";
+import { minDevice } from "../../../assets/styles/breakpoints";
+import { blue200, placeholder } from "../../../assets/styles/colors";
+
+const SubtitleWrapper = styled(Wrapper)`
+  color: ${placeholder};
```
franiakat: I'd rename this color, similar to what @bazelajoanna did with blue200 and so on). Now seeing its usage we see the name should be different, as it looks a bit odd here. ;)

franiakat (commented):
Looking good! 😎 Just some minor details to consider.

bazelajoanna (commented):
🥇 👩‍🎓

src/components/atoms/ProblemSubtitle/styles.tsx
evulins: Done :)

franiakat (approved):
sqash

PR #37, opened 2022-01-04
Add library and functionality to generate pdf

evulins:
1. Generating and downloading pdf using html2canvas and jsPDF libraries.
2. Downloading existing pdf file using Link from react-router-dom

For tests using this solution https://github.com/cypress-io/cypress-example-recipes/tree/master/examples/testing-dom__download

Review of other solutions related to generating pdf - https://start-up.house/en/blog/articles/converting-html-to-pdf


README.md
```diff
@@ -66,7 +66,11 @@ Open [http://localhost:3000](http://localhost:3000) to view it in the browser.
 
 ### `yarn test`
 
-Launches the test runner in the interactive watch mode.
+Launches the test runner in the interactive watch mode (using [Testing Library](https://testing-library.com/) and [Jest](https://jestjs.io/))
```
franiakat: Awesome, that you are supplementing the docs 😻  Thous this particular command starts the tests not in the watch mode `watchAll=false` : >

README.md
```diff
@@ -66,7 +66,11 @@ Open [http://localhost:3000](http://localhost:3000) to view it in the browser.
...
+
+### `yarn cyprsess:open`
```
franiakat:
```suggestion
### `yarn cypress:open`
```

README.md
```diff
@@ -66,7 +66,11 @@ Open [http://localhost:3000](http://localhost:3000) to view it in the browser.
...
+
+Launches the test runner in the interactive watch mode (using [Cypress](https://www.cypress.io/))
```
franiakat: There is also the regular `yarn cypress` that does not open the browser, you can add it as well.

cypress/integration/problem_detailed_page.js
```diff
@@ -19,4 +19,29 @@ describe("Problem Detailed Page", () => {
     cy.get(`a[href="/problems/1"]`).first().click();
     cy.contains("Pobierz treść");
   });
+
+  const path = require("path");
```
franiakat: I'd place it directly after the imports.

cypress/integration/problem_detailed_page.js
```diff
@@ -19,4 +19,29 @@ describe("Problem Detailed Page", () => {
...
+
+  it("Initiate download pdf file containing problem details and verify it", () => {
```
franiakat: Let's use one convention, either use lower case here or change the old cypress tests titles, please ;)
Also, `Initiates` ;)

cypress/integration/problem_detailed_page.js
```diff
@@ -19,4 +19,29 @@ describe("Problem Detailed Page", () => {
...
+      .should("be.visible")
+      .click();
+    cy.wait(1000);
+    const downloadsFolder = Cypress.config("downloadsFolder");
+    cy.readFile(path.join(downloadsFolder, "Ciągi matematyczne.pdf")).should(
+      "exist"
+    );
+  });
+
+  it("Initiate download pdf file containing additional resources and verify it", () => {
```
franiakat: `initiates`

cypress/integration/problem_detailed_page.js
```diff
@@ -19,4 +19,29 @@ describe("Problem Detailed Page", () => {
     cy.get(`a[href="/problems/1"]`).first().click();
     cy.contains("Pobierz treść");
   });
+
+  const path = require("path");
+
+  it("Initiate download pdf file containing problem details and verify it", () => {
+    cy.get(`a[href="/problems/1"]`).first().click();
+
+    cy.get('[class*="Button-sc-1v3p2bo-0 kuukQO"]')
```
franiakat: You can use `contains` as well, would be more human friendly then:
```suggestion
    cy.contains('Pobierz treść').should("be.visible").click();
```

cypress/integration/problem_detailed_page.js
```diff
@@ -19,4 +19,29 @@ describe("Problem Detailed Page", () => {
...
+    cy.wait(1000);
+    const downloadsFolder = Cypress.config("downloadsFolder");
+    cy.readFile(path.join(downloadsFolder, "Ciągi matematyczne.pdf")).should(
+      "exist"
+    );
+  });
+
+  it("Initiate download pdf file containing additional resources and verify it", () => {
+    cy.get(`a[href="/problems/1"]`).first().click();
+    cy.get('[class*="StyledButtonLink"]').should("be.visible").click();
```
franiakat:
```suggestion
    cy.contains('additional resources').should("be.visible").click();
```

public/locales/pl/translation.json
```diff
@@ -0,0 +1,43 @@
+{
```
franiakat: Why is this whole file added? 🤔  It's already on main.

franiakat: Is there a correct base branch picked for this feature branch? Not in the PR, as I see it is main here, but actually the branch you were starting locally at first?

src/assets/styles/global-styles.js
```diff
@@ -91,6 +91,7 @@ ${fonts}
     border-collapse: collapse;
     border-spacing: 0;
   }
+}
```
franiakat, line 94: 🐒  woopsie, this is weird change :D

src/components/pages/ProblemDetailedPage/index.tsx
```diff
@@ -45,15 +52,45 @@ const ProblemDetailedPage: FC = () => {
...
+
+    const canvas = await html2canvas(element);
+    const imgData = canvas.toDataURL("image/png");
+    const pdf = new jsPDF("p", "mm");
+
+    let position = 0;
+
+    const imgProperties = pdf.getImageProperties(imgData);
+    const pdfWidth = pdf.internal.pageSize.getWidth() - 20;
+    const pageHeight = 295;
```
franiakat: Where is this number taken from? :)

src/components/pages/ProblemDetailedPage/index.tsx
```diff
@@ -45,15 +52,45 @@ const ProblemDetailedPage: FC = () => {
...
+    const element = printRef.current as HTMLDivElement;
+
+    const canvas = await html2canvas(element);
+    const imgData = canvas.toDataURL("image/png");
+    const pdf = new jsPDF("p", "mm");
+
+    let position = 0;
+
+    const imgProperties = pdf.getImageProperties(imgData);
+    const pdfWidth = pdf.internal.pageSize.getWidth() - 20;
```
franiakat: Can't we just use `width` instead of `getWidth()`? : >

franiakat (changes requested):
🐵  AWESOME JOB with the research and implementation, Ewu! 😸  We had just a few comments to offer our perspective.

cypress/integration/problem_detailed_page.js
evulins: I am not sure about that  - because "additional resources" at design has the name of the file. In that case it will not always be the same text.

src/assets/styles/global-styles.js
evulins: I am not sure what happened here :D

src/components/pages/ProblemDetailedPage/index.tsx
evulins: This is  more or less the hight of A4 format

src/components/pages/ProblemDetailedPage/index.tsx
evulins: I don't think so -> https://github.com/parallax/jsPDF/issues/1832

public/locales/pl/translation.json
evulins: Once when i was  pulling main branch I used merge not rebase maybe because of that my PR looks like that

cypress/integration/problem_detailed_page.js
franiakat: But it matches what we have in our mock data for the problem with id=1, unless we change mock data in handlers, it won't change. :)

public/locales/pl/translation.json
franiakat: Hm, interesting, maybe  :D

src/components/pages/ProblemDetailedPage/index.tsx
franiakat: I thought there is correlation as 297 is A4 height ;)

src/components/pages/ProblemDetailedPage/index.tsx
franiakat:
```
    const pdfWidth = pdf.internal.pageSize.width - 20;
```
Works for me as well, I noticed that, as I hovered over to see the interface of the `pdf.internal` ;)

evulins: It this any particular reason for changing it? I don't mind to change it, i am just curious. :)

src/components/pages/ProblemDetailedPage/index.tsx
evulins: I can change it to exact a4 height in our case those 3 points doesn't change a thing.

public/locales/pl/translation.json
evulins: Once more time I rebase and push it with force

src/components/pages/ProblemDetailedPage/index.tsx
franiakat: No need to ;)

src/components/pages/ProblemDetailedPage/index.tsx
franiakat: If there is just a key that gets what I won't I would use it over having to invoke the function that gets the same value, it just looks a bit simpler for me and doesn't make me think, opposed to `getWidth()` ;) But you can treat it as a neat pick, wouldn't "request change" for that for sure :D

src/App.test.tsx
```diff
@@ -85,4 +85,15 @@ describe("<App />", () => {
 
     expect(await screen.findByText("problemExtension")).toBeInTheDocument();
   });
+
+  it.only("shows problem detailed page after clicking on suggestion ", async () => {
```
franiakat: leftover `only` ;)

src/components/pages/ProblemDetailedPage/index.tsx
```diff
@@ -65,7 +99,7 @@ const ProblemDetailedPage: FC = () => {
           </S.ProblemCreationDetails>
           <S.ProblemCreationDetails>{formattedDate}</S.ProblemCreationDetails>
         </S.ProblemCreationDetailsWrapper>
-        <S.Section>
+        <S.Section data-html2canvas-ignore="true">
```
franiakat, line 103: 👏

franiakat (approved):
Great work, ready to meet with everyone else on main :)
sqash

PR #41, opened 2022-01-11
Add link to whole search result tile

bazelajoanna:
Make the wole SearchResultTile clickable and able to navigate to ProblemDetailedPage.

### DIFF
--- a/cypress/integration/problem_detailed_page.js
+++ b/cypress/integration/problem_detailed_page.js
@@ -19,10 +19,4 @@ describe("Problem Detailed Page", () => {
     cy.get(`a[href="/problems/1"]`).first().click();
     cy.contains("Pobierz treść");
   });
-
-  it("shows problem detailed page after clicking on search result tile on mobile and tablet", () => {
-    cy.viewport("iphone-6");
-    cy.get('[class*="SearchResultTile"]').first().click();
-    cy.contains("Pobierz treść");
-  });
 });
--- a/cypress/integration/problem_detailed_page.js
+++ b/cypress/integration/problem_detailed_page.js
@@ -19,4 +19,10 @@ describe("Problem Detailed Page", () => {
     cy.get(`a[href="/problems/1"]`).first().click();
     cy.contains("Pobierz treść");
   });
+
+  it("shows problem detailed page after clicking on search result tile on mobile and tablet", () => {
+    cy.viewport("iphone-6");
+    cy.get('[class*="SearchResultTile"]').first().click();
+    cy.contains("Pobierz treść");
+  });
 });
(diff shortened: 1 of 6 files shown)


cypress/integration/problem_detailed_page.js
```diff
@@ -19,4 +19,10 @@ describe("Problem Detailed Page", () => {
     cy.get(`a[href="/problems/1"]`).first().click();
     cy.contains("Pobierz treść");
   });
+
+  it("shows problem detailed page after clicking on search result tile on mobile and tablet", () => {
```
evulins: I am just wondering if this test should be here or in the new file for problems page? What do you think?

src/components/molecules/SearchResultTile/index.tsx
```diff
@@ -19,9 +18,9 @@ const SearchResultTile: FC<ProblemSummaryProps> = ({
   return (
     <S.SearchResultTileWrapper>
       <S.DetailsWrapper>
-        <Link to={`/problems/${id}`}>
-          <S.Title>{title}</S.Title>
-        </Link>
+        <S.Title>
+          <S.StyledTitleLink to={`/problems/${id}`}>{title}</S.StyledTitleLink>
+        </S.Title>
```
evulins, line 23: 👏

evulins (commented):
Works like a charm! :)

cypress/integration/problem_detailed_page.js
franiakat: There.

src/App.test.tsx
```diff
@@ -85,4 +85,14 @@ describe("<App />", () => {
...
+
+  it("doesn't show problem detailed page after clicking on search result tile not clickable item on large dekstop", async () => {
+    testRenderer(<App />);
+    const author = (await screen.findByText(
+      "Maria Konopnicka"
+    )) as HTMLInputElement;
+    userEvent.click(author);
+
+    expect(await screen.findByText(/searchResults/)).toBeInTheDocument();
+  });
```
franiakat: I'd test it in cypress, specifying exactly the tool : >

cypress/integration/problem_detailed_page.js
bazelajoanna: Right! 🐈

src/App.test.tsx
bazelajoanna: I've moved it to cypress, by default it tests on screen with dimension 1000x600, so I haven't specified the tool.

franiakat (approved):
👏 nice Asia, I like the way it's implemented, thanks for the cypress tests as well!

PR #42, opened 2022-01-13
Fix style in mail icon 

bazelajoanna:
Fix size of the e-mail icon and spacing between it and the e-mail address.


evulins (approved):
🎊

franiakat (commented):
@bazelajoanna I think you can remove the pdf files, right? :)

franiakat (approved):

PR #44, opened 2022-01-14
Add login form

evulins:
Create a form that allows to log in, so that an authenticated user is able to see protected page (Strefa Moderatora).
Add JWT tokens and store them in LocalStorage. https://jwt.io/introduction

Checkbox functionality will be provided in separate pull request


public/locales/pl/translation.json
```diff
@@ -39,5 +39,21 @@
     "contact": "Kontakt",
     "email": "myslaceklasy@email.com",
     "copyrights": "© Copyrights 2020."
+  },
+  "loginPage": {
+    "loginHeader": "Zaloguj się",
+    "loginSubtitle": "*Opcja logowania dostępna wyłącznie dla naszych Moderatorów.",
+    "loginErrorMessage": "Nieprawidłowy login lub hasło",
+    "emailErrorMessage": "Błędny email",
+    "passwordErrorMessage": "Błędne hasło",
```
bazelajoanna, line 48: These descriptions are unused because we had problems with adding useTranslation inside `validations.tsx`

src/components/pages/LoginPage/index.test.tsx
```diff
@@ -0,0 +1,70 @@
...
+  });
+
+  it("shows password after clicking on an eye icon", async () => {
+    testRenderer(<LoginPage />);
+    const passwordInput = (await screen.findByPlaceholderText(
+      "inputPasswordPlaceholder"
+    )) as HTMLInputElement;
+
+    userEvent.type(passwordInput, "test123");
+    const eyeIcon = screen.findByAltText("showPasswordIconAlt");
```
evulins: What is the difference between this test and the one above?  In the ened we don't use here eyeIcon. Maybe it is a way to check if this icon is working in our test environment?

src/components/pages/LoginPage/index.test.tsx
```diff
@@ -0,0 +1,70 @@
...
+  it("displays error message when inputted password didn't pass validation", async () => {
+    testRenderer(<LoginPage />);
+    const passwordInput = (await screen.findByPlaceholderText(
+      "inputPasswordPlaceholder"
+    )) as HTMLInputElement;
+
+    userEvent.type(passwordInput, "test");
+    userEvent.type(passwordInput, "{enter}");
+
+    expect(screen.queryByText(/Błędne hasło/i)).not.toBeInTheDocument();
```
evulins: As @franiakat once wrote: toBe or notBe?

evulins: I think, If we type wrong password error message should be displayed

src/components/pages/LoginPage/styles.tsx
```diff
@@ -0,0 +1,262 @@
...
+import { ReactComponent as Eye } from "../../../assets/img/icons/eye.svg";
+
+import {
+  white,
+  blue100,
+  blue300,
+  primaryBackground,
+  primaryText,
+  grey100,
+  grey200,
```
evulins, line 21: Grey-ish colors needs to be checked. One of them has smaller number and in fact is darker than the other 🤓

src/index.tsx
```diff
@@ -5,12 +5,12 @@ import { BrowserRouter } from "react-router-dom";
 import { ApolloProvider } from "@apollo/client";
 
 import App from "./App";
+import "./i18n";
 import { client } from "./apolloClient";
-import GlobalStyle from "./assets/styles/global-styles";
+import { worker } from "./mocks/browser";
```
evulins, line 10: Eslint suggested this change. We can double check if it's working when we don't use required.

src/components/pages/LoginPage/styles.tsx
```diff
@@ -0,0 +1,262 @@
...
+  height: 5.254rem;
+  position: absolute;
+  right: 0;
+  @media ${minDevice.desktopTablet} {
+    width: 16.343rem;
+    height: 11.231rem;
+  }
+
+  @media ${minDevice.desktopL} {
+    width: 25.36rem
```
evulins: Semicolon is missing here

src/components/pages/LoginPage/styles.tsx
```diff
@@ -0,0 +1,262 @@
...
+    height: 22.08rem;
+  }
+
+  @media ${minDevice.desktopXL} {
+    width: 31.7rem
+    height: 27.6rem
```
evulins: Semicolons are missing here

src/components/pages/LoginPage/styles.tsx
```diff
@@ -0,0 +1,262 @@
...
+    background-image: url(${LeftTopBubbles});
+    background-size: contain;
+    background-repeat: no-repeat;
+    width: 10rem;
+    height: 9rem;
+    position: absolute;
+    left: 0;
+  }
+  @media ${minDevice.desktopL} {
+    width: 9.947rem
```
evulins: Semicolon is missing here

src/components/pages/LoginPage/index.tsx
```diff
@@ -0,0 +1,119 @@
...
+                  <S.InputWrapper>
+                    <S.Label>
+                      {t("password")}
+                      <S.ShownPasswordButton
+                        onClick={togglePassword}
+                        aria-pressed="false"
+                        aria-label={t("showPasswordIconAlt")}
+                        role="switch"
+                      >
+                        {passwordShown ? <S.EyeIcon /> : <S.EyeOffIcon />}
```
evulins: When I'm clicking on eyeIcon and don't have any value inputted then i got an error "Nieprawidłowy login lub hasło", this is strange🤔

evulins (commented):
Amazing work :)  I have small questions :)

src/components/pages/LoginPage/index.test.tsx
bazelajoanna: This test is not finished 🙈  - we want to check if the password is visible after clicking on eyeIcon. Tomorrow we could think how to check it, because `findByText` is not helpful here.

src/components/pages/LoginPage/index.test.tsx
bazelajoanna: You're 💯 % right. But toBe doesn't work :( We'll look at this tomorrow as well.

src/components/pages/LoginPage/index.tsx
bazelajoanna: Weird.. we'll look at this 👩‍🎓

src/components/pages/LoginPage/index.test.tsx
bazelajoanna: I've moved this test to cypress, now it works :)

src/components/pages/LoginPage/index.test.tsx
bazelajoanna: This test now checks if the input has type `text` instead of `password`. And it works 👍

src/components/pages/LoginPage/index.tsx
evulins: Fixed :)

cypress/integration/login_page.js
```diff
@@ -0,0 +1,20 @@
+describe("Problem Detailed Page", () => {
+  beforeEach(() => {
+    cy.visit("http://localhost:3000/");
+    cy.get('a[href*="/moderator"]').click();
+  });
+
+  it("displays error message when inputted email and password didn't pass validation", () => {
```
franiakat:
```suggestion
  it("displays error message when inputted email and password didn't pass client-side validation", () => {
```

cypress/integration/login_page.js
```diff
@@ -0,0 +1,20 @@
...
+    cy.get('input[name="email"]').type("test");
+    cy.get('input[name="password"]').type("123");
+
+    cy.contains("Błędny email");
+  });
+
+  it("displays error message when inputted password didn't pass validation", () => {
```
franiakat: When is the passoword invalid? It should be for sure checked for presence, but nothing else, as proper credentials validation happens on the backend. ;)

franiakat: Oh, now I see, so this sort of validation will be useful when the user is creating a password, but not when logging in, it can be too informative for an attacker. ;)

src/components/pages/LoginPage/index.tsx
```diff
@@ -0,0 +1,122 @@
...
+      <S.LoginFormWrapper>
+        <S.H4>{t("loginHeader")}</S.H4>
+        <S.InfoText>{t("loginSubtitle")}</S.InfoText>
+        <S.StyledLink to="/">{t("learnMore")}</S.StyledLink>
+        {error && (
+          <S.LoginErrorMessage>{t("loginErrorMessage")}</S.LoginErrorMessage>
+        )}
+        <Form
+          onSubmit={onSubmit}
+          validate={formInputsValidation}
```
franiakat:
```
          validate={(values) =>
            formInputsValidation(values, t("emailErrorMessage"))
          }
```

src/components/atoms/ErrorMessage/index.tsx
```diff
@@ -0,0 +1,12 @@
+import styled from "styled-components";
+import { errorMessage } from "../../../assets/styles/colors";
+
+const ErrorMessage = styled.span`
+  display: flex;
+  font-size: 1.4rem;
+  color: ${errorMessage};
```
franiakat: I am not a fan of this name for a colour, maybe sth about `alert`? or simply `error`, so would be just skipping message. ;)

src/components/pages/LoginPage/validations.tsx
```diff
@@ -0,0 +1,32 @@
+import { FormInputsType } from "./interface";
+
+const formInputsValidation = (values: {
```
franiakat: You could pass two translations needed for the error message. You are invoking this function in a component, where you are already using the `t` function, so I do with sth along those lines:

const formInputsValidation = (
  values: {
    email?: string;
    password?: string;
  },
  wrongEmailLabel: string

src/components/pages/LoginPage/validations.tsx
```diff
@@ -0,0 +1,32 @@
...
+  const { email, password } = values;
+
+  const errors = {} as FormInputsType;
+
+  if (!email || (email && !validateEmail(email))) {
+    errors.email = " Błędny email";
+  }
+  if (!password || (password && !validatePassword(password))) {
+    errors.password = " Błędne hasło";
+  }
```
franiakat: I'd keep only the condition for its presence.

src/components/pages/LoginPage/index.tsx
```diff
@@ -0,0 +1,122 @@
...
+                    )}
+                  </S.InputWrapper>
+                )}
+              </Field>
+              <Field name="password">
+                {({ input, meta }) => (
+                  <S.InputWrapper>
+                    <S.Label>{t("password")}</S.Label>
+                    <S.Input
+                      type={passwordShown ? "text" : "password"}
```
franiakat, line 77: nice!

src/App.tsx
```diff
@@ -1,26 +1,44 @@
...
-        {/* <Route path="/" element {<LandingPage/>}> */}
-        <Route path="/" element={<ProblemsPage />} />
-        <Route path="/problems/:problemId" element={<ProblemDetailedPage />} />
-        <Route path="/about" element={<AboutMethodPage />} />
-        <Route path="/moderator" element={<Moderator />} />
-      </Routes>
-      <Footer />
-    </Suspense>
+    <AuthProvider>
+      <Suspense fallback={<Loader />}>
```
franiakat: Also works with <Suspense fallback={null}> I think ;)

src/auth/AuthProvider.tsx
```diff
@@ -0,0 +1,48 @@
...
+  const navigate = useNavigate();
+
+  const userData = localStorage.getItem("user");
+  const parsedUserData = userData && JSON.parse(userData);
+
+  useEffect(() => {
+    if (!!parsedUserData == true) {
+      setUser(parsedUserData);
+    }
+  }, []);
```
franiakat, line 22: What if user is removed from local storage when the app is already mounted? There is nothing listening to this change - we can sort it out in separate PR though. :)

src/auth/AuthProvider.tsx
```diff
@@ -0,0 +1,48 @@
...
+  const [login, { loading, error }] = useMutation<Login>(LOGIN);
+
+  const signin = (email: string, password: string) => {
+    login({
+      variables: { email: email, password: password },
+    })
+      .then((data) => {
+        const user = data.data?.login;
+        setUser(user);
+        localStorage.setItem("user", JSON.stringify(user));
```
franiakat: I would set only JWT token value in local storage, so we would have this pair:
```
localStorage.setItem("user", JSON.stringify({token: user.token}));
```

franiakat (changes requested):
Wow, what a beast! I really like your tests and good job with implementation as well! I have a few thoughts, I must admit I was barely checking styling and I skipped checking Checkbox whatsoever. : >

cypress/integration/login_page.js
bazelajoanna: I've changed it to: `it displays error message when inputted email didn't pass client-side validation` as we test here only email validation.

cypress/integration/login_page.js
bazelajoanna: Ok, I removed those conditions :)

src/components/pages/LoginPage/validations.tsx
bazelajoanna: Nice! Thank you :) I've added also `wrongPasswordLabel`, as we have two different messages :)

src/components/pages/LoginPage/index.tsx
bazelajoanna: @evulins 👸

src/App.tsx
bazelajoanna: Could you please describe it more clearly? 🤔 I don't know what we should change here ;)

src/components/pages/LoginPage/index.tsx
bazelajoanna: Same here, I've added also `t("passwordErrorMessage")`

src/components/pages/LoginPage/validations.tsx
franiakat: Great! Although we don't want to have a` wrong passoword label`, but actually only `missing password label `, as discussed in context of validation. ;)

src/App.tsx
franiakat: Haha, sorry command C didn't work here!
This should work and the loader is not needed:
```
      <Suspense fallback={null}>
```

franiakat (approved):
sqash

PR #46, opened 2022-01-28
Add checkbox

evulins:
Add checkbox styled component and provide functionality for "remember me" button in LoginPage. This functionality uses local storage and JWT to recognise user and store user data for specified period of time.


src/auth/AuthProvider.tsx
```diff
@@ -2,35 +2,45 @@ import { useMutation } from "@apollo/client";
...
 export const AuthProvider: FC<AuthProps> = ({ children }) => {
-  const [user, setUser] = useState<User>();
+  const [user, setUser] = useState<{ token: string; expiry: number }>();
 
   const navigate = useNavigate();
-
-  const userData = localStorage.getItem("user");
-  const parsedUserData = userData && JSON.parse(userData);
+  const userData = getUserDataFromStorage("user");
+  const noUserData = !!userData;
```
franiakat: Here you are making a boolean value out of the object, so no negation. ;) It's rather like creating a flag:
```suggestion
  const isUserDataPresent = !!userData;
```

src/auth/AuthProvider.tsx
```diff
@@ -2,35 +2,45 @@ import { useMutation } from "@apollo/client";
...
-  const signin = (email: string, password: string) => {
+  const signin = (email: string, password: string, checked: boolean) => {
     login({
       variables: { email: email, password: password },
     })
       .then((data) => {
         const user = data.data?.login;
-        setUser(user);
-        localStorage.setItem("user", JSON.stringify({ token: user?.token }));
+        user && setUser({ token: user?.token, expiry: 123 });
```
franiakat, line 34: Do you think we want to store and actual expiry in the store? Probably it won't be useful, so I am wondering whether the `setUser` function should be more of a `setUserToken`, that needs only this one piece of information.

src/auth/interface.ts
```diff
@@ -2,12 +2,22 @@ import { ApolloError } from "@apollo/client";
...
+
+export interface UserTokenWithExpiry {
+  token: string;
+  expiry: number;
+}
+
+export interface UserTokenWithOptionalExpiry {
+  token: string;
+  expiry?: number;
+}
```
franiakat, line 23: I wasn't sure whether both were actually needed. ;)

src/components/pages/LoginPage/interface.ts
```diff
@@ -16,6 +16,7 @@ export interface User {
   id: number;
   email: string;
   token: string;
+  expiry?: number;
```
franiakat, line 19: Yeah, here as well the user interface makes me think : ) Which things are needed inside the context :)

src/auth/AuthProvider.tsx
```diff
@@ -2,35 +2,45 @@ import { useMutation } from "@apollo/client";
...
+        if (user && !checked) {
+          setUserDataInStorage("user", user.token, oneDayInMS);
+          window.onbeforeunload = function () {
+            localStorage.clear();
```
franiakat, line 38: We can one day go back to it and see how we could leverage custom hook here. :)

franiakat (changes requested):

bazelajoanna (approved):
Wonderful!

PR #47, opened 2022-02-01
Fix wrong credentials notification

bazelajoanna:
Fix following bug in LoginPage by updating context.

When user types email and password that pass validation but are not mocked in database, the notification about wrong email or password appears. When they go to e.q. main page and then return to Moderator page the email and password inputs has been refreshed, but the notification stays.


franiakat (approved):
🎉

